PR #8, opened 2021-12-30
feat/dropdown: add option to view multiple cell values in a dropdown & separated options for features view

dayesouza:
- add checkbox to view the dropdown when the value is an array
- add dropdown with onChange handler so the user can pass an arrayAsDropdown boolean prop to render it this way
- add onCellDropdownSelect prop to ArqueroDetailsList so the user can pass a function to be the dropdown's onChange


.yarn/versions/32c88654.yml
```diff
@@ -0,0 +1,3 @@
+releases:
+  "@data-wrangling-components/react": patch
```
natoverse, line 2: I would probably call this "minor" since it involves adding new functionality to the public API.
- patch: small fix/update with no visible API changes
- minor: API additions that are not breaking changes
- major: breaking changes

packages/react/src/ArqueroDetailsList/types.ts
```diff
@@ -2,15 +2,24 @@
...
 import ColumnTable from 'arquero/dist/types/table/column-table'
 
 export type ColumnRenderFunction = (
 	item?: any,
 	index?: number,
 	column?: IColumn,
+	arrayAsDropdown?: boolean,
 ) => any
 
+export type DropdownOptionSelect =
```
natoverse, line 15: I think this select handler would be more useful if we could pass along more information about the cell. A normal dropdown is just a list of items to choose, but in this case you might want to know what column and row it was as well. IDropdownOption (extends ISelectableOption) has a generic type T and a `data` field to hold whatever you'd like - that could be a way to do this by passing through the column object and rowIndex.

packages/react/src/ArqueroDetailsList/renderers/SmartArrayCell.tsx
```diff
@@ -33,9 +34,12 @@ export const SmartArrayCell: React.FC<RichCellProps> = memo(
 				<Case condition={values.length <= cellWidth}>
 					<SparkbarCell {...props} />
 				</Case>
-				<Case condition={values.length > cellWidth}>
+				<Case condition={values.length > cellWidth && !arrayAsDropdown}>
```
natoverse, line 37: I think if we are going to have `arrayAsDropdown` as a separate feature rather than automatic if `smartCells` is turned on, then we should probably override _any_ array cell to be a dropdown instead of a chart. In other words, either you get a plan text array cell, a smart cell with charts, or a dropdown.

natoverse (changes requested):

packages/react/src/ArqueroDetailsList/hooks/useCellDropdownSelectHandler.ts
```diff
@@ -0,0 +1,19 @@
...
+import { DropdownOptionSelect } from '..'
+
+export function useCellDropdownSelectHandler(
+	clickable: boolean,
+	onOptionSelect?: DropdownOptionSelect,
+): DropdownOptionSelect | undefined {
+	return useMemo(
+		() =>
+			clickable
+				? (evt?, column?) => onOptionSelect && onOptionSelect(evt, column)
```
natoverse, line 15: Make sure to pass along the index arg too

packages/react/src/ArqueroDetailsList/util/index.ts
```diff
@@ -9,6 +9,16 @@ export function getValue(item: any, column?: IColumn): any {
 	return column?.fieldName && item[column.fieldName]
 }
 
+export function getDropdownValue(item: any, column?: IColumn): any {
+	const itens = column?.fieldName && item[column.fieldName]
+	return itens.map((value: any) => {
+		return {
+			key: value,
```
natoverse, line 16: I see React errors about duplicate keys - we may want to add the index to the key to ensure it stays unique

PR #12, opened 2022-01-04
Arquero table header

gaudyb:
- Moved table header into a new component inside react package.
- For the header props, I added a table ColumnTable, showColumnCount, showRowCount and allowDownload props.
- Fixed all eslint errors.
- Preinstall error fixed using codespaces.


.yarnrc.yml
```diff
@@ -1,6 +1,3 @@
-changesetIgnorePatterns:
```
natoverse, line 1: Can you add this back in?

natoverse (approved):

.yarnrc.yml
gaudyb: Yes sorry, fixed

PR #30, opened 2022-01-10
Transformation steps (new verbs)

gaudyb:
- New verbs inside core -> verbs folder (rollup, ungroup, unorder, dedupe, impute, fetch)
- New jest tests inside core engine for testing new verbs
- Modified csv files inside webapp/public/data without a bug character at the beginning of each file


.yarn/versions/b14843cf.yml
```diff
@@ -0,0 +1,4 @@
+releases:
+  "@data-wrangling-components/core": major
```
natoverse, line 2: I would actually release this as "minor", since you are adding new API functionality but not breaking backwards-compatibility

packages/core/src/engine/verbs/fetch.ts
```diff
@@ -0,0 +1,23 @@
...
+import ColumnTable from 'arquero/dist/types/table/column-table'
+
+/**
+ * Executes an arquero loadCSV
+ * @param step
+ * @param store
+ * @returns
+ */
+export async function fetch(
+	url: string,
```
natoverse, line 16: I would use the same signature as the other verbs here (step, store) so that they are chainable in a pipeline.  You should have a FetchStep with FetchArgs, where the FetchArgs contains a require url field and an optional delimiter field

packages/core/src/engine/verbs/index.ts
```diff
@@ -22,17 +23,21 @@ import { lookup } from './lookup'
...
+import { rollup } from './rollup'
 import { sample } from './sample'
 import { select } from './select'
 import { spread } from './spread'
+import { ungroup } from './ungroup'
 import { union } from './union'
+import { unorder } from './unorder'
 import { unroll } from './unroll'
 
 const verbs: Record<string, StepFunction> = {
```
natoverse, line 35: Add fetch into this list too

packages/core/src/types.ts
```diff
@@ -161,10 +161,12 @@ export enum Verb {
 	Aggregate = 'aggregate',
 	Bin = 'bin',
 	Binarize = 'binarize',
+	Dedupe = 'dedupe',
 	Derive = 'derive',
 	Fill = 'fill',
 	Concat = 'concat',
 	Except = 'except',
+	Impute = 'impute',
```
natoverse, line 169: Add Fetch = 'fetch'

natoverse (changes requested):

packages/react/src/controls/dropdowns/FieldAggregateOperationDropdown.tsx
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) Microsoft. All rights reserved.
  * Licensed under the MIT license. See LICENSE file in the project.
  */
-import { FieldAggregateOperation } from '@data-wrangling-components/core'
+import { FieldAggregateRollupOperation } from '@data-wrangling-components/core'
```
natoverse, line 5: You can keep the name of this as `FieldAggregateOperation` since aggregate/rollup basically mean the same thing

natoverse (approved):

PR #46, opened 2022-01-18
New verbs UI

gaudyb:


.yarn/versions/de4a9a51.yml
```diff
@@ -0,0 +1,4 @@
+releases:
+  "@data-wrangling-components/core": patch
+  "@data-wrangling-components/react": major
```
natoverse, line 3: Let's set this to "minor" for now, since you are adding new verbs to the API but not disrupting the previous API (i.e., it is still backwards compatible)

packages/react/src/verbs/Ungroup/Ungroup.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * Licensed under the MIT license. See LICENSE file in the project.
+ */
+import React, { memo } from 'react'
+import { LeftAlignedRow, VerbContainer } from '../../common'
+import { StepComponentProps } from '../../types'
+
+/**
+ * Provides inputs for an ungroup step.
+ */
+export const Ungroup: React.FC<StepComponentProps> = memo(function Ungroup({
```
natoverse, line 12: Ungroup and Unroll could probably both be collapsed into something like a "NoParameters" component that just returns null. This is like the SetOperation - each set verb (difference, intersect, etc.) returns the same SetOperation component in selectstepComponent.

natoverse (approved):
Added suggestions but they aren't critical. Please make sure the lint is good, and if you want to do the other changes then go ahead and merge when ready.

PR #64, opened 2022-02-01
Fix some known bugs

monik182:


packages/core/src/types.ts
```diff
@@ -104,6 +104,7 @@ export enum SetOp {
 export enum SortDirection {
 	Ascending = 'asc',
 	Descending = 'desc',
+	Unsorted = 'unsorted',
```
natoverse, line 107: Could this just be represented by an undefined SortDirection in state rather than an enum value?

packages/react/src/ArqueroTableHeader/ArqueroTableHeader.tsx
```diff
@@ -41,34 +47,35 @@ export const ArqueroTableHeader: React.FC<ArqueroTableHeaderProps> = memo(
...
-	display: flex;
-	justify-content: space-around;
-	align-items: center;
 	background-color: ${({ theme }) => theme.application().accent().hex()};
+	position: relative;
+	padding: 0 5px;
+	box-sizing: border-box;
+	display: inline-flex;
+	align-items: center;
+	justify-content: space-around;
```
natoverse, line 73: I think with the two-command-bar options this would now look better with `justify-content: space-between;` to push the command bards to the edges with the metadata centered. That would also require the farCommands to default to flex-end so they are right-justified.

natoverse (changes requested):

PR #79, opened 2022-02-14
Thematic Update; Move Core, Utilities to ESM Builds

darthtrevino:
* Use common Essex tsconfigs 
* Migrate dwc/core, dwc/utils to esm
* Remove .yarn/cache


package.json
```diff
@@ -4,31 +4,31 @@
...
+		"start": "yarn workspaces foreach -piv run start",
+		"deploy": "yarn workspaces foreach -piv run deploy",
+		"lint": "essex lint --fix",
 		"git_is_clean": "essex git-is-clean",
 		"prettify:": "essex prettify",
-		"test:": "NODE_OPTIONS=--experimental-vm-modules yarn jest",
-		"ci": "run-s build: lint: test: git_is_clean",
+		"jest": "NODE_OPTIONS=--experimental-vm-modules yarn jest",
+		"ci": "run-s build lint jest git_is_clean",
 		"publish:": "yarn workspaces foreach --exclude '@data-wrangling-components/project' -pv npm publish --tolerate-republish --access public",
```
natoverse: colon

package.json
```diff
@@ -4,31 +4,31 @@
...
+		"clean": "yarn workspaces foreach -pv run clean",
+		"build": "yarn workspaces foreach -pvt run build",
+		"start": "yarn workspaces foreach -piv run start",
+		"deploy": "yarn workspaces foreach -piv run deploy",
+		"lint": "essex lint --fix",
 		"git_is_clean": "essex git-is-clean",
 		"prettify:": "essex prettify",
-		"test:": "NODE_OPTIONS=--experimental-vm-modules yarn jest",
-		"ci": "run-s build: lint: test: git_is_clean",
+		"jest": "NODE_OPTIONS=--experimental-vm-modules yarn jest",
```
natoverse: seems like we should still have a root-level `test`, even if it is an alias for `jest` - this would still avoid recursion, right?

darthtrevino: Because there are no package-specific tests, making "test' just run "jest" is fine

PR #89, opened 2022-02-18
react router implementation

gaudyb:
BrowserRouter implemented for Webapp navigation.


packages/webapp/src/components/App/Layout.tsx
```diff
@@ -0,0 +1,34 @@
...
+`
+const Container = styled.div`
+	margin: 0;
+	height: 100vh;
+	display: flex;
+	flex-flow: column;
+`
+
+const Content = styled.div`
+	margin: 0 20px 20px 20px;
```
natoverse, line 30: Set `margin: 0;` so the container has no effect on the child page (which we use no margin on)

natoverse (changes requested):

natoverse (approved):

PR #124, opened 2022-03-05
Fix a copy-paste-o?

dreness:
This fixes what appears to be an unfortunate documentation error in the code example for `unfold`.

In terms of value per character, I'd say this is not quite as good as flipping a sign or moving a decimal point, but not far behind either :)

Very cool project, by the way!


darthtrevino:
Good catch, thank you!

PR #136, opened 2022-03-10
Feature/save load

monik182:


natoverse:
A couple of comments on the ux/eventing
- Open should be left of save
- The open icon would be more clear if it looks more like a file (like "OpenFile"). The plain arrow looks more like an upload, which isn't quite accurate
- In the dropdowns Open CSV table should be first since that is the most common
- When a table is opened it should immediately display on the page - this may need a prop on PrepareData like "selectedTable"
- The hard-coded test tables should be removed
- Drag/drop of zip project files doesn't work (it works fine with the open button)
- We'll need to figure out something to deal with pre-computed outputs. If I create a pipeline and save the zip, when reloading it runs, which results in the previously computed output as an input (e.g., table-bin.csv) and also a copy in derived (table-bin).

PR #160, opened 2022-03-21
Add lookup label in lookup step

dayesouza:


natoverse (approved):
Looks great - another approach, which we use elsewhere, would be to default to using the verb as the label, which in this case would mean you don't even need to pass in a label. But for these things it makes sense to have options to override the default anyway.

PR #169, opened 2022-03-23
Use Graph Processing Model

darthtrevino:


javascript/core/src/steps/types.ts
```diff
@@ -0,0 +1,100 @@
...
+
+	/**
+	 * The bound inputs
+	 */
+	inputs: Record<string, { node: string; output?: string }>
+
+	/**
+	 * The observed outputs to record.
+	 */
+	pinnedOutputs: Array<{ output?: string; name: string }>
```
natoverse, line 68: For serialized pipelines it would be nice if the inputs/outputs on the Step were symmetrically named and structured

javascript/core/src/verbs/stepFunctions/simpleSteps.ts
```diff
@@ -0,0 +1,133 @@
...
+export const eraseStep: TableStep<EraseArgs> = (
+	input: ColumnTable,
+	{ value, column }: EraseArgs,
+) => {
+	const func = escape((d: any) => (d[column] === value ? undefined : d[column]))
+	return input.derive({ [column]: func })
+}
+
+export const fillStep: TableStep<FillArgs> = (input, { value, to }) => {
+	const fn = (_d: any, $: any) => $.value
```
natoverse, line 64: This doesn't looks like a valid arquero expression

javascript/core/src/steps/types.ts
darthtrevino: We could improve symmetry by using `outputs` instead of `pinnedOutputs`, but the structure won't match entirely.  Full symmetry is a bit of a misnomer because it's not a port-binding (like in the inputs area) but a publication to the TableStore. 

So inputs would be `Record<string, Binding>` and outputs would be `Record<string,string>`

javascript/core/src/verbs/stepFunctions/simpleSteps.ts
```diff
@@ -0,0 +1,133 @@
...
+	SelectArgs,
+	SpreadArgs,
+	UnrollArgs,
+	WindowArgs,
+} from '../types/domainTypes.js'
+import { deriveBoolean } from '../util/expressions.js'
+import { compareAll, singleExpression } from '../util/index.js'
+import type { ExprFunctionMap } from '../util/types.js'
+
+export const aggregateStep: TableStep<AggregateArgs> = (
```
natoverse, line 34: I think I prefer having the steps broken into individual files. At the expense of more file management it is easier to fine and open files, especially when they are grouped into "types" like "simple steps" that may not be immediately apparent

javascript/core/src/verbs/stepFunctions/simpleSteps.ts
```diff
@@ -0,0 +1,133 @@
...
+	input,
+	{ groupby, column, operation, to },
+) => {
+	const expr = singleExpression(column, operation)
+	return input.groupby(groupby).rollup({ [to]: expr })
+}
+
+export const binarizeStep: TableStep<BinarizeArgs> = (
```
natoverse, line 42: How do we differentiate sync and async steps now with the elimination of promises? I.e., fetch returns a promise, but the rest of these would be async only within an observable context, right?

natoverse: Is the StepNode factory how promises are consistently chained?

darthtrevino: The execution steps can be either `async` or `sync`. We avoid `await` in the node implementations so that if data is available synchronously it will compute without an extra `tick()`. If you look for `handleMaybeAsync`, you'll see where that happens.

javascript/core/src/verbs/stepFunctions/simpleSteps.ts
darthtrevino: Yeah I went back and forth on this. A bunch of tiny files was kind of annoying but I get the argument against clumping them. I like separating the stepFunctions from the node factories (it makes it easier to test), but this means that you don't have a full, canonical "verb list".

Would breaking up simpleSteps.ts into separate files satisfy your concern here?

javascript/core/src/verbs/stepFunctions/simpleSteps.ts
darthtrevino: The `stepNodeFactory` creates a Node Factory out of a step function.  The chaining has to be explicit with this model. I'd take a look at `graph.spec.ts` to see that in action

javascript/core/src/verbs/stepFunctions/simpleSteps.ts
natoverse: I think so - just a flat list of verb files seems the easiest to follow

darthtrevino: sure thing, I'm fine with one-file per verb

javascript/core/src/steps/types.ts
darthtrevino: In my latest push this is a record<string,string> I'm not sure it's an improvement, but it rhymes with the inputs object. One caveat is that if you're using the default node output you have to use the key `default`: 

e.g. 
```js
{
   outputs: { default: 'table-1' }
}
```

I also updated the graph composer so that it will implicitly use the default output of the previous step as the default input if no inputs are defined. It feels a little weird, but it may be okay for this use case. Since the UI is going to be constructing these steps I'm not sure it's worth removing that explicit wiring.

PR #180, opened 2022-03-30
Feature/convert improvements

gaudyb:


javascript/core/src/engine/verbs/convert.ts
```diff
@@ -26,19 +27,46 @@ const doConvert = wrapColumnStep<ConvertArgs>(
...
+					: parseTime(value)
+			}
 			case ParseType.Integer:
 				return op.parse_int(value, radix)
 			case ParseType.Decimal:
 				return op.parse_float(value)
+			case ParseType.String: {
+				if (
+					typeof value === 'string' &&
+					value.trim().toLowerCase() === 'undefined'
```
natoverse, line 59: I think lodash isNil could simplify this check

javascript/core/src/engine/verbs/convert.ts
```diff
@@ -26,19 +27,46 @@ const doConvert = wrapColumnStep<ConvertArgs>(
...
+				)
+					return undefined
+
+				if (typeof value === 'string' && value.trim().toLowerCase() === 'null')
```
natoverse, line 63: isNil should cover null and undefined at once. I think we also have an isEmpty that checks for empty strings

javascript/core/src/engine/verbs/convert.ts
```diff
@@ -26,19 +27,46 @@ const doConvert = wrapColumnStep<ConvertArgs>(
...
+					? isoParse(value)
+					: parseTime(value)
+			}
 			case ParseType.Integer:
 				return op.parse_int(value, radix)
 			case ParseType.Decimal:
 				return op.parse_float(value)
+			case ParseType.String: {
+				if (
+					typeof value === 'string' &&
```
natoverse, line 58: We have a determineType util that centralizes type checking and returns an enum you can use for switching

natoverse (changes requested):
Couple of minor comments on utils to you. There are also some lint things that should be fixed (ignore the files and commands ones, those have been around and should be fixed on main so everyone gets them)

javascript/core/src/engine/verbs/convert.ts
gaudyb: the same as above happens here I think

javascript/core/src/engine/verbs/convert.ts
natoverse: Ah, I see. Ok, let's leave that as is. It might change with the new data types rules, but I'd rather go through all of those in one sweep for consistency anyway.

natoverse (approved):

PR #216, opened 2022-04-21
Task/updates

darthtrevino:


javascript/react-verbs/docs/react-verbs.api.md
```diff
@@ -317,7 +328,7 @@ export type StepChangeFunction = (step: Step) => void;
 // Warning: (ae-missing-release-tag) "StepComponentProps" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
 //
 // @public (undocumented)
-export interface StepComponentProps extends StepDependent {
+export interface StepComponentProps<T extends object = any> extends StepDependent<T> {
```
natoverse, line 331: Shouldn't this extend `unknown` by default?

javascript/react-verbs/src/common/hooks.ts
```diff
@@ -40,15 +46,21 @@ export function useHandleDropdownChange(
 	)
 }
 
+// #endregion
+
+// #region Textfield Change Handler
+
+export type TextfieldChangeHandler = (
```
natoverse, line 53: This is super persnickety, but fluent caps the component as 'TextField' so it might be a little more clear

javascript/react/src/TableTransformModal/hooks/useModalStyles.ts
```diff
@@ -16,7 +16,7 @@ import { useMemo } from 'react'
 export function useModalStyles(
 	styles?: IStyleFunctionOrObject<IModalStyleProps, IModalStyles>,
 	includeGuidance = false,
-	hasSelectedStep = false,
+	_hasSelectedStep = false,
```
natoverse, line 19: It looks like this should just come off the function

javascript/react-controls/src/controls/ColumnSpread.tsx
```diff
@@ -23,15 +23,12 @@ export const ColumnSpread: React.FC<ColumnSpreadProps> = memo(
...
+				opt && onChange?.(opt)
 			},
 			[onChange],
 		)
 
-		const handleDeleteClick = useCallback(
-			() => onDelete && onDelete(),
-			[onDelete],
-		)
+		const handleDeleteClick = useCallback(() => onDelete?.(), [onDelete])
```
natoverse, line 31: this could be omitted entirely since it is optional on the IconButton anyway

javascript/react-verbs/src/common/hooks.ts
```diff
@@ -10,15 +10,21 @@ import type {
...
-} from '../types.js'
+import type { StepChangeFunction } from '../types.js'
+
+// #region Dropdown Change Handler
+
+export type DropdownChangeHandler = (
+	event: React.FormEvent<HTMLDivElement>,
+	option?: IDropdownOption,
+	index?: number,
+) => void
```
natoverse, line 27: we have tons of inlined function typings for fluent handlers - maybe a central types/fluent.ts to contain these?

javascript/react-verbs/src/verbs/Aggregate/Aggregate.tsx
```diff
@@ -20,56 +23,50 @@ import type { StepComponentProps } from '../../types.js'
...
+		)
 
-	const handleGroupColumnChange = useHandleDropdownChange(
-		internal,
-		'args.groupby',
-		onChange,
-	)
+		const handleGroupColumnChange = useHandleDropdownChange(
+			step,
+			'args.groupby',
```
natoverse, line 36: Curious your thoughts on these. `set` breaks typings, but also makes it really convenient to reuse these handlers. Any cleaner ideas?

javascript/react-verbs/src/verbs/Convert/Convert.tsx
```diff
@@ -31,105 +31,100 @@ import { getColumnType } from '../shared/TypingFunction/TypingFunction.js'
...
-	const internal = useMemo(() => step as ConvertStep, [step])
-	const tbl = useLoadTable(
-		input || internal.input[NodeInput.Source]?.node,
-		table,
-		store,
-	)
-	const [inputColumnDate, setInputColumnDate] = useState<boolean>()
+export const Convert: React.FC<StepComponentProps<ConvertArgs>> = memo(
+	function Convert({ step, store, table, onChange, input }) {
+		const tbl = useLoadTable(
```
natoverse, line 36: We could probably abstract all of these with an HOC that injects the resolved table?

javascript/react-verbs/src/verbs/Convert/Convert.tsx
```diff
@@ -31,105 +31,100 @@ import { getColumnType } from '../shared/TypingFunction/TypingFunction.js'
...
-		},
-		[step, onChange],
-	)
+		const handleComboBoxChange = useCallback(
+			(
+				_e: any,
+				option: IComboBoxOption | undefined,
+				_index: number | undefined,
+				value: string | undefined,
+			) => {
```
natoverse, line 61: fluent ComboBoxChangeHandler?

javascript/react-verbs/src/verbs/Spread/Spread.tsx
```diff
@@ -20,80 +19,73 @@ import type { StepComponentProps } from '../../types.js'
...
  */
-export const Spread: React.FC<StepComponentProps> = memo(function Spread({
-	step,
-	store,
-	table,
-	onChange,
-	input,
-}) {
-	const internal = useMemo(() => step as SpreadStep, [step])
+export const Spread: React.FC<StepComponentProps<SpreadArgs>> = memo(
```
natoverse, line 22: Have you gotten rid of the Step aliases entirely in favor of specifying the args type (e.g., `SpreadStep = Step<SpreadArgs>`)?

javascript/react/src/ColumnTransformModal/hooks/useHandleRunClick.tsx
```diff
@@ -7,12 +7,12 @@ import type { Step } from '@data-wrangling-components/core'
...
-	internal: Step | undefined,
-	onTransformRequested?: (internal: Step) => void,
+	step: Step | undefined,
+	onTransformRequested?: (step: Step) => void,
 ): () => void {
 	return useCallback(() => {
-		if (internal) {
-			onTransformRequested && onTransformRequested(internal)
+		if (step) {
+			onTransformRequested?.(step)
```
natoverse, line 15: `step && onTransformRequested..(step)`?

javascript/webapp/src/pages/DebugPage/ControlBar.tsx
```diff
@@ -85,7 +85,7 @@ export const ControlBar: React.FC<ControlBarProps> = memo(function ControlBar({
 			if (!first) return
 			updateFileCollection([first])
 			const spec = await loadSpec(first)
-			onSelectSpecification && onSelectSpecification(spec as any)
+			onSelectSpecification?.(spec as any)
```
natoverse, line 88: I can't remember why these are cast to `any`

natoverse (changes requested):
Much cleaner - probably a good reduction in loc

javascript/react-verbs/src/verbs/Spread/Spread.tsx
darthtrevino: Yeah, just to tighten things down. I'm not sure the XStep types were useful

javascript/react-verbs/docs/react-verbs.api.md
darthtrevino: Yep, unknown is reasonable

javascript/react-verbs/src/common/hooks.ts
darthtrevino: Hmm, maybe, but I'm kind of a fan of colocating types with where they're primarily provided from. I'll have a think

javascript/react-verbs/src/verbs/Aggregate/Aggregate.tsx
darthtrevino: What if we had a callback function for handling the update instead? This would make these functions more generically useful. We could guarantee immutability with Immer (https://immerjs.github.io/immer/update-patterns), which is what react-toolkit uses to preserve immutability with state changes.

javascript/react-verbs/src/verbs/Convert/Convert.tsx
darthtrevino: I'd rather use a hook that took a props object to streamline this. I'm personally not a fan of HOCs

javascript/react-verbs/src/verbs/Convert/Convert.tsx
darthtrevino: Possibly, it's only used once, and the usage isn't super generic. If we moved to a callback-based update pattern I could see this

javascript/react-verbs/src/verbs/Spread/Spread.tsx
darthtrevino: Yes, in an effort to streamline the exported surface area of Core. I didn't find much utility in the Step aliases.

javascript/react-verbs/src/verbs/Convert/Convert.tsx
darthtrevino: Eh, a HOC is cleaning this up pretty nicely

javascript/react-controls/src/controls/ColumnSpread.tsx
darthtrevino: We still have to like...call the delete handler right?

javascript/react-verbs/src/common/hooks.ts
darthtrevino: It doesn't make a ton of sense in context here, but I like this idea for an `essex-js-toolkit` package

natoverse: maybe so - I guess this app doesn't have a _ton_ of them directly, but in aggregate across our apps we do. It's actually probably something that should be pushed up to fluent though - they're just inconsistent about typings on function handlers.

javascript/react-verbs/src/verbs/Convert/Convert.tsx
natoverse: isn't this the standard onChange definition for fluent ComboBox? we define handler types for the rest of these (i.e., previous fluent typings comment)

javascript/react-verbs/src/verbs/Aggregate/Aggregate.tsx
natoverse: I don't think I'm tracking. This is a callback function, it's just a pattern we use so frequently that it's abstracted to a reusable dropdown-change-property-setter. What from immerjs would be different and still keep out the boilerplate this avoids?

javascript/react-verbs/src/verbs/Convert/Convert.tsx
darthtrevino: Yes, but the update mattern may be different (will it use index ever? is the value fallback always consistent?)

javascript/react-verbs/src/verbs/Aggregate/Aggregate.tsx
darthtrevino: Something like this: 
```
const handleGroupColumnChange = useHandleDropdownChange(
	step,
	(args, option) => args.groupby = option?.key // vs.  'args.groupby'; we can get type safety with the callback
	onChange,
)
```

darthtrevino: Immer would replace cloneDeep, so the handler could be written: 

```ts
export function useHandleDropdownChange<T extends object | void | unknown>(
	step: Step<T>,
	updateFn: (step: Step<T>, update: string) => void,
	onChange?: StepChangeFunction<T>,
): DropdownChangeHandler {
	return useCallback(
		(_event, option) => {
			onChange?.(produce(step,  draft => updateFn(draft, option?.key)))
		},
		[step, path, onChange],
	)
}
```

natoverse: Got it - great idea

javascript/react-verbs/src/verbs/Convert/Convert.tsx
natoverse: Why is this different [than this](https://github.com/microsoft/data-wrangling-components/blob/9e15859ff84b0522c9cf46607f4473cfdcf7e250/javascript/react-verbs/src/common/hooks.ts#L23)?

The callback interface needs to be compliant with the fluent callback either way - you can name the variables whatever you want.

PR #218, opened 2022-04-23
Date comparisons for binarize/filter verbs

gaudyb:


javascript/core/src/verbs/util/compare.ts
```diff
@@ -132,3 +139,24 @@ function compareBooleans(
 			throw new Error(`Unsupported boolean comparison operator: [${operator}]`)
 	}
 }
+
+function compareDates(
+	left: Date,
+	right: Date,
+	operator: DateComparisonOperator,
+): 1 | 0 {
+	switch (operator) {
```
natoverse, line 148: can all of these just use the valueOf number?

javascript/core/src/verbs/util/compare.ts
```diff
@@ -132,3 +139,24 @@ function compareBooleans(
...
+		case DateComparisonOperator.Equals:
+			return left.toLocaleDateString() === right.toLocaleDateString() ? 1 : 0
+		case DateComparisonOperator.NotEqual:
+			return left.toLocaleDateString() !== right.toLocaleDateString() ? 1 : 0
+		case DateComparisonOperator.After:
+		case DateComparisonOperator.AfterToday:
+			return left > right ? 1 : 0
```
natoverse, line 155: I think it would be more clear if you created `right = new Date()` here in the comparison - otherwise, someone using the compareDates function might not know that they need to create their own today date before calling this function.

javascript/core/src/verbs/util/expressions.ts
```diff
@@ -34,8 +36,10 @@ export function compareAll(
 		// cannot be satisfied
 		const comparisons = criteria.map(filter => {
 			const { value, operator, type } = filter
-			const right =
+			let right =
 				type === FilterCompareType.Column ? d[`${value.toString()}`]! : value
+			if(operator as DateComparisonOperator === DateComparisonOperator.AfterToday || operator as DateComparisonOperator === DateComparisonOperator.BeforeToday)
```
natoverse, line 41: moving the `new Date()` in the compareDates function would get rid of this special case here and put it in the type-specific comparison spot

javascript/react-verbs/src/verbs/shared/FilterFunction/FilterFunction.tsx
```diff
@@ -126,17 +147,22 @@ export const FilterFunction: React.FC<FilterFunctionProps> = memo(
...
-						table={table}
-						filter={columnFilter}
-						disabled={isEmpty}
-						label={suppressLabels ? undefined : 'Comparison value'}
-						placeholder={placeholder}
-						text={criterion.value ? `${criterion.value}` : undefined}
-						onChange={handleComboBoxChange}
-						styles={narrowDropdownStyles}
-					/>
+					{type === DataType.Date ? (
```
natoverse, line 150: This one is kind of tricky. If we just show the calendar for dates, we lose the ability to compare to another column

natoverse (dismissed):

natoverse (approved):

PR #231, opened 2022-05-02
Table naming

natoverse:
- fixes the issue with conflated table ids and names (just uses id)
- refactors hooks/components in a couple of places
- cleans up styles


javascript/core/src/steps/readStep.ts
```diff
@@ -2,8 +2,6 @@
  * Copyright (c) Microsoft. All rights reserved.
  * Licensed under the MIT license. See LICENSE file in the project.
  */
-import { v4 as uuid } from 'uuid'
```
darthtrevino, line 5: We can probably remove the UUID dep

darthtrevino (approved):

PR #250, opened 2022-05-17
Feature/array data type

gaudyb:


javascript/react/src/verbs/Convert.base.tsx
```diff
@@ -54,8 +64,24 @@ export const ConvertBase: React.FC<
...
 	)
 
 	return <VerbForm inputs={inputs} step={step} onChange={onChange} />
 })
+
+function isInputColumnArray(columnsMetadata: ColumnMetadata[], columns: string[]){
+	let arrayFlag = false
+	let index = 0
+
+	while(!arrayFlag && index < columns.length){
```
darthtrevino, line 77: We can simplify this by removing the column lookup and using Array.some (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

e.g.
```ts
return columnsMetadata.some((metadata) = metadata.type === DataType.Array)
```

I don't thinking looking up the column name is doing much

javascript/react/src/hooks/useColumnsMetadata.ts
```diff
@@ -0,0 +1,33 @@
...
+
+        if (!table || !col) {
+			type = DataType.Unknown
+		}
+        else
+            type = columnType(table, col)
+
+        result.push({columnName: col, type: type})
+    })
+	return useMemo(() => result, [table, filter, result])
```
darthtrevino, line 27: Lines 13-26 should be inside the useMemo, or else the memoization isn't saving any work

javascript/react/src/hooks/useColumnsMetadata.ts
```diff
@@ -0,0 +1,33 @@
...
+import { columnType, DataType } from '@essex/arquero'
+import type ColumnTable from 'arquero/dist/types/table/column-table'
+import { useMemo } from 'react'
+
+export function useColumnsMetadata(
+	table: ColumnTable | undefined,
+	filter?: (name: string) => boolean,
+): ColumnMetadata[] {
+    const columns = table?.columnNames(filter) || []
+    const result: ColumnMetadata[] = []
```
darthtrevino, line 14: let's use Array.map instead: 

```ts
const result: ColumnMetadata[] = columns.map(col => {
// return metadata object
})
```

javascript/react/src/hooks/useColumnsMetadata.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    columns.forEach(col => {
+        let type: DataType
+
+        if (!table || !col) {
+			type = DataType.Unknown
+		}
+        else
+            type = columnType(table, col)
```
darthtrevino, line 23: let's either use braces or a ternary: 

e.g.
```ts
const type: DataType = (!table || !col) ? DataType.Unknown : columnType(table,col)
```

darthtrevino (changes requested):

javascript/react/src/verbs/Convert.base.tsx
gaudyb: It is needed by column, because the component receives all column names and we need to enable the array delimiter just if one column with array datatype was selected into the multidropdown.

darthtrevino: gotcha, maybe we can filter down the metadata list first then?

PR #251, opened 2022-05-18
reset verb args when changing input table

dayesouza:


javascript/react/src/hooks/common.ts
```diff
@@ -67,3 +67,19 @@ export function useFormattedColumnArg(): (
 		return args
 	}, [])
 }
+
+export function useResetTableRelatedArgs(): (stepArgs: unknown) => object {
+	return useCallback((stepArgs: unknown) => {
+		const args = stepArgs as Record<string, unknown>
+		Object.keys(args).forEach(key => {
+			if (typeof args[key] === 'string') {
```
darthtrevino, line 75: Are there any other arg-types that we need to consider, numbers or booleans?

darthtrevino (approved):

javascript/react/src/hooks/common.ts
dayesouza: Changed numbers to be reseted to 0 (undefined did nothing to number step component) and boolean to undefined

PR #259, opened 2022-05-27
hook rename, added update workflow listener

monik182:


javascript/react/src/hooks/uploads.ts
```diff
@@ -106,10 +106,10 @@ export function useHandleZipUpload(
 		},
 		[csvHandler, jsonHandler, onUpdateTables],
 	)
-	return useHandleOnUploadClick(['.zip'], handler)
+	return useUploadFileHandler(['.zip'], handler)
 }
 
-export function useHandleOnUploadClick(
+export function useUploadFileHandler(
```
darthtrevino, line 112: Let's rename this `useHandleFileUpload` to keep the naming convention used in `useHandleCsvUpload` and `useHandleJsonUpload`

darthtrevino (changes requested):
Small rename comment; looks good otherwise

darthtrevino (approved):

PR #262, opened 2022-06-02
Fix/bug

monik182:


javascript/react/src/hooks/downloads.ts
```diff
@@ -56,7 +57,12 @@ export function useDownloadZip(
 		}
 
 		if (tables.length) {
-			const files = tables
+			const map = new Map<string, number>()
+			const renamedTables = tables.map(t => ({
+				...t,
+				name: renameDuplicatedFileName(map, t.name || t.id || 'table.csv'),
```
darthtrevino, line 63: will we run into an issue if there are multiple files without names?

javascript/react/src/hooks/downloads.ts
```diff
@@ -56,7 +57,12 @@ export function useDownloadZip(
...
+			}))
+			const files = renamedTables
 				.map(table => tableFile(table))
 				.filter(f => f !== null)
```
darthtrevino, line 67: `!== null` is better expressed as `!= null` 

The reasoning here is that !== is explicitly checking for the null value, while != null will capture null and undefined.

javascript/utilities/src/utils/files.ts
```diff
@@ -162,34 +162,45 @@ export async function getDataURL(file: BaseFile): Promise<string> {
...
-		if (count > 1) {
-			const ext = extension(name)
-			name = `${name.replace(`.${ext}`, '')} (${count}).${ext}`
-		}
-		return createBaseFile(file, { name })
-	})
+	return values.some(val => val > 1)
+}
+
+function cleanName(name: string): string {
```
darthtrevino, line 199: rename to `cleanFileName`

javascript/utilities/src/utils/files.ts
```diff
@@ -162,34 +162,45 @@ export async function getDataURL(file: BaseFile): Promise<string> {
...
+	name: string,
+): string {
+	const next = (map.get(name) || 0) + 1
+	map.set(name, next)
+	const ext = extension(name)
+	return next > 1 ? `${name.replace(`.${ext}`, '')} (${next}).${ext}` : name
+}
+
+function hasDuplicatedNames(names: string[]): boolean {
+	const fileNames: Record<string, number> = {}
```
darthtrevino, line 188: Using a Set<string> makes more sense here. 
```
let uniqueNames = new Set<string>(names)
return uniqueNames.size< names.length
```

darthtrevino (changes requested):

javascript/react/src/hooks/downloads.ts
darthtrevino: I see, the renameDupliactedFileName will resolve the name for us

monik182: I checked, and we might. So far, when a csv file is uploaded, the id of the tableContainer object created for that file is the file name. I tried changing it to be a random string, but it broke the table select, as it showed the id of the selected file and not its name as it currently does.

PR #263, opened 2022-06-03
Spread update

natoverse:
Adds new parameters to the SPREAD verb:
- delimiter: optional delimiter to pre-split string cell values into an array before spreading. This addresses a common use case with csv data files that have arrays in cells that can't be represented any other way
- onehot: the spread operation becomes a ONEHOT encoding into new columns for each cell value instead of using the default array index approach of arquero. This covers many more machine learning use cases than the default spread behavior.

Note that the react components have been updated to add the new params and _remove_ the "to" arg. This is an uncommon arg, but supports custom behavior in the arquero spread impl. It can still be used in specifications, but if user control is desired, then a separate UX will need to be used for these special cases.


javascript/react/src/verbDescriptions/SpreadDescription.tsx
```diff
@@ -6,22 +6,39 @@ import type { SpreadArgs } from '@data-wrangling-components/core'
...
 import { VerbDescription } from '../verbForm/VerbDescription.js'
 
 export const SpreadDescription: React.FC<StepDescriptionProps<SpreadArgs>> =
 	memo(function SpreadDescription(props) {
+		const {
+			step: { args },
+		} = props
+		const sub = createRowEntries(
+			args.columns,
+			c => ({
```
darthtrevino, line 19: streamline to: `value => ({value})`

darthtrevino (approved):

PR #265, opened 2022-06-03
Python/graph

andresmor-ms:
* Compatibility with latest schema changes
* Fix differences on results with javascript
* Bug Fixing


python/data-wrangling-components/data_wrangling_components/engine/verbs/join.py
```diff
@@ -3,52 +3,63 @@
...
+    other = other.table
 
-    if len(args.on) > 1:
-        left_column = args.on[0]
-        right_column = args.on[1] if len(args.on) > 0 else None
+    if on is not None and len(on) > 1:
+        left_column = on[0]
+        right_column = on[1]
 
         output = input_table.merge(
```
natoverse, line 48: What is the difference between merge and join in pandas?

natoverse: Does the `sort` keyword affect the results in a way that might match the JS default?

natoverse: Does the `sort` keyword affect the results in a way that might match the JS default?

natoverse (approved):

python/data-wrangling-components/data_wrangling_components/engine/verbs/join.py
andresmor-ms: I tried to use the sort keyword but it didn't help much :( 

Join uses merge internally, join combines tables using the index of the dataframe and merge uses columns, I have always used merge since i find it easier, however in this scenario they're pretty much interchangeable

PR #275, opened 2022-06-18
updated onehot to allow multiple columns

monik182:


javascript/core/src/verbs/onehot.ts
```diff
@@ -22,22 +22,29 @@ export interface OnehotArgs extends InputColumnArgs {
...
-
-	const args = distinct.sort().reduce((acc, cur) => {
-		acc[prefix ? `${prefix}${cur}` : cur] = escape((d: any) =>
-			d[column] === null ? null : d[column] === cur ? 1 : 0,
-		)
-		return acc
-	}, {} as Record<string, ExprObject>)
+	let args = {}
+
+	columns.forEach(column => {
```
natoverse, line 29: This would be cleaner if you used `const args = columns.reduce(...)` instead of reassigning `args` each time

javascript/core/src/verbs/onehot.ts
```diff
@@ -22,22 +22,27 @@ export interface OnehotArgs extends InputColumnArgs {
...
+
+		const colArgs = distinct.sort().reduce((acc, cur) => {
+			acc[prefix ? `${prefix}${column}_${cur}` : `${column}_${cur}`] = escape(
+				(d: any) => (d[column] === null ? null : d[column] === cur ? 1 : 0),
+			)
+			return acc
+		}, {} as Record<string, ExprObject>)
+
+		return { ...acc, ...colArgs }
+	}, {})
```
natoverse, line 45: Are you getting type warnings here? Should be `{} as Record<string, ExprObject>`

andresmor-ms (approved):

python/data-wrangling-components/data_wrangling_components/engine/verbs/onehot.py
```diff
@@ -3,16 +3,34 @@
 # Licensed under the MIT license. See LICENSE file in the project.
 #
 
+from typing import Dict, List
```
andresmor-ms, line 6: Please make sure that you use `black` and `isort` 

`poetry run black data_wrangling_components/`
`poetry run isort data_wrangling_components/`

javascript/react/src/verbs/OneHot.base.tsx
```diff
@@ -2,29 +2,73 @@
...
+		] as FormInput<OnehotArgs>[]
+	}, [step, columns, prefixInputs])
+
+	const onStepChange = useCallback(
+		(s: Step<OnehotArgs>) => {
+			const step = cloneDeep(s)
+			if (!step.args.prefixes) {
+				step.args.prefixes = {}
+			} else {
+				const prefixes = setInitialPrefixes(step)
```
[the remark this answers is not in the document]
monik182: I tested it and I'm sure the prefix is not being reset to the default value

I added the 2 columns for onehot and deleted the prefix for one of them 
Before
image
After
image

javascript/core/src/verbs/onehot.ts
monik182: I already changed it to the reduce, I don't know why is not updated here.
This is the file in the branch I have the changes to merge: https://github.com/microsoft/data-wrangling-components/blob/feature/onehot-multiple-columns/javascript/core/src/verbs/onehot.ts

javascript/core/src/verbs/onehot.ts
```diff
@@ -22,22 +22,27 @@ export interface OnehotArgs extends InputColumnArgs {
...
+			.get('distinct', 0) as any[]
+
+		const colArgs = distinct.sort().reduce((acc, cur) => {
+			acc[prefix ? `${prefix}${column}_${cur}` : `${column}_${cur}`] = escape(
+				(d: any) => (d[column] === null ? null : d[column] === cur ? 1 : 0),
+			)
+			return acc
+		}, {} as Record<string, ExprObject>)
+
+		return { ...acc, ...colArgs }
```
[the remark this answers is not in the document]
monik182: I already fixed this too: https://github.com/microsoft/data-wrangling-components/blob/feature/onehot-multiple-columns/javascript/core/src/verbs/onehot.ts

PR #276, opened 2022-06-29
boolean xnor implementation ts and python impl

gaudyb:


javascript/core/src/verbs/util/boolean-logic.ts
```diff
@@ -87,25 +89,28 @@ export function and(comparisons: (1 | 0 | null)[]): 1 | 0 | null {
  * @returns
  */
 export function xor(comparisons: (1 | 0 | null)[]): 1 | 0 | null {
```
natoverse, line 91: Should we replace the bulk of this function with the even/odd logic to simplify?

javascript/core/src/verbs/util/boolean-logic.ts
```diff
@@ -156,3 +161,20 @@ export function nand(comparisons: (1 | 0 | null)[]): 1 | 0 | null {
 	}
 	return null
 }
+
+/**
+ * Logical XNOR
+ * @param comparisons
+ * @returns
+ */
+ export function xnor(comparisons: (1 | 0 | null)[]): 1 | 0 | null {
```
natoverse, line 170: Please double-check that the code comments on each of these functions matches with our latest and most accurate descriptions of the logic.

natoverse (approved):
I added a couple of comments that I think would be work addressing, but fine to merge once you get there

javascript/core/src/verbs/util/boolean-logic.ts
```diff
@@ -87,25 +89,28 @@ export function and(comparisons: (1 | 0 | null)[]): 1 | 0 | null {
...
-			// shortcut more than one true immediately
-			return 0
-		}
-		if (comparisons[i] === null) {
-			nulls++
-		}
+	let xor = null
+
+	if(comparisons.length > 0){
+		if(comparisons[0] === null)
```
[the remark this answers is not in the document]
gaudyb: it's not redundant because the loop starts at 1

PR #286, opened 2022-07-20
Feature/merge onehot columns

gaudyb:


javascript/core/src/verbs/merge.ts
```diff
@@ -49,7 +69,47 @@ export const mergeStep: ColumnTableStep<MergeArgs> = (
...
+		return unhot
+			? copyTable.derive({ [to]: func })
+			: input.derive({ [to]: func })
+
+	return unhot
+		? copyTable.derive({ [to]: func }).select(not(columns))
+		: input.derive({ [to]: func }).select(not(columns))
+}
+
+function unhotOperation(
```
natoverse, line 82: Rather than mutating the inputTable within this function, you should create a new output table and return it, so the function is "pure"

javascript/core/src/verbs/merge.ts
```diff
@@ -26,13 +26,33 @@ export interface MergeArgs extends InputColumnListArgs, OutputColumnArgs {
...
+		strategy,
+		to,
+		delimiter = '',
+		unhot = false,
+		prefix = '',
+		keepOriginalColumns = false,
+	},
 ) => {
-	const isSameDataTypeFlag: boolean = isSameDataType(input, columns)
+	const copyTable: ColumnTable = fromJSON(input.toJSON())
```
natoverse, line 46: I don't think you will need to use this copyTable line if you make unhotOperation a pure function and don't mutate the input table

javascript/core/src/verbs/merge.ts
```diff
@@ -26,13 +26,33 @@ export interface MergeArgs extends InputColumnListArgs, OutputColumnArgs {
...
+
+	if (unhot) {
+		unhotOperation(copyTable, columns, prefix)
+	}
+
+	const isSameDataTypeFlag: boolean = isSameDataType(
+		unhot ? copyTable : input,
```
natoverse, line 53: You could get rid of all of these `unhot ? copyTable : input` if you create a single intermediate table earlier and just use it from there on out. For example:

`const working = unhot ? unhotOperation(input...) : input`

Then just use `working` directly afterwards instead of the ternary operator

javascript/react/src/verbs/Merge.base.tsx
```diff
@@ -35,6 +35,25 @@ export const MergeBase: React.FC<
 				current: step.args.delimiter,
 				onChange: (s, val) => (s.args.delimiter = val as string),
 			},
+			{
```
natoverse, line 38: On main I added the ability to collapse extra params into an "advanced" expand/collapse section. I think a bunch of these could go in there to reduce clutter on the ux, particularly related to unhot stuff that will be less commonly used. You can pass `advanced: true` to the control generator functions (check out VerbForm). For this verb I would suggest the following get pushed into advanced: unhot, keepOriginalColumns, prefix.

javascript/core/src/verbs/merge.ts
```diff
@@ -26,13 +26,33 @@ export interface MergeArgs extends InputColumnListArgs, OutputColumnArgs {
 	 * If it is not supplied, the values are just mashed together.
 	 */
 	delimiter?: string
+	unhot?: boolean
```
natoverse, line 29: Please add some additional docs for these params:
- unhot: Input columns have been one-hot encoded, the merge should extract the original value from the column name
- prefix: When using `unhot`, this is a shared column name prefix to remove from the value
- keepOriginalColumns: Merging eliminates the original source columns by default, collapsing into one. Set this to `true` to return the sources.

natoverse (changes requested):

javascript/core/src/verbs/merge.ts
```diff
@@ -49,7 +62,46 @@ export const mergeStep: ColumnTableStep<MergeArgs> = (
...
+
+	return tempTable.derive({ [to]: func }).select(not(columns))
+}
+
+function unhotOperation(
+	input: ColumnTable,
+	columns: string[],
+	prefix: string,
+): ColumnTable {
+	const inputTable: ColumnTable = fromJSON(input.toJSON())
```
natoverse, line 75: You should be able to use `inputTable.reify()` here, which should perform better

natoverse (approved):

PR #296, opened 2022-08-17
Move types

natoverse:
Moves a few things from core and arquero packages into the schema package to start centralizing all of the types that drive schema generation.


javascript/core/src/index.ts
```diff
@@ -4,7 +4,5 @@
  */
 export * from './engine/index.js'
 export * from './primitives.js'
-export * from './steps/index.js'
```
darthtrevino, line 7: Do we want to re-export the verb types and core-related types from `@datashaper/schema` or will clients be expected to use types from that package?

darthtrevino (approved):

javascript/core/src/index.ts
natoverse: I guess we could go either way. I figured everyone would just include schema as a dep, but that's not strictly necessary if it would feel similar as a reexport

PR #302, opened 2022-08-19
Bug/date format pattern

gaudyb:


javascript/react/src/verbForm/VerbForm.tsx
```diff
@@ -110,6 +110,8 @@ export interface ComboBoxFormInput<T>
 	 */
 	current: string | undefined
 
+	allowFreeForm?: boolean
```
natoverse, line 113: Can you fix the case on this so it matches the prop on Combobox exactly? (`allowFreeform`). I think that will be less potentially confusing.

natoverse (changes requested):

natoverse (approved):

PR #315, opened 2022-08-31
Refactor/steps history view

monik182:


dayesouza (approved):

natoverse:
These updates look really good! A few layout comments:
- Does 100% height work ok? 90% is going to be weird I think
- Make the icon smaller and print the work "History" on the button
- Add a close button on the panel - it isn't clear that re-clicking the history icon is required to close, and the top-right is the standard fluent guidance for close buttons
- I would remove the faint border around each step panel and rely on the single horizontal divider instead. Too many boxes will end up looking cluttered

PR #316, opened 2022-09-01
Webpage refactor

natoverse:
- Removes the app navigation entirely - the debug page should be secondary at this point and only really used by us, so the nav link is back in the settings panel to hide it a bit
- Cleans up the webapp a bit, moving components into page folders to better organize them (fully flat was a bit much with so many components being page-specific)
- Fixes up the styles and layout for a bunch of the verbs to make sure they all have the same maximum width. This is preparation for usage in a fixed-width history panel


dayesouza (approved):

javascript/react/src/controls/SortInstruction.tsx
```diff
@@ -14,6 +16,13 @@ import { Container, icons } from './SortInstruction.styles.js'
 import type { SortInstructionProps } from './SortInstruction.types.js'
 import { TableColumnDropdown } from './TableColumnDropdown.js'
 
+const narrowerStyles = merge({}, dropdownStyles, {
```
darthtrevino, line 19: Is the empty object even necessary here?

javascript/react/src/verbs/shared/FilterFunction.tsx
```diff
@@ -208,9 +202,18 @@ export const FilterFunction: React.FC<FilterFunctionProps> = memo(
 
 		return (
 			<Container>
-				<SideBySide>
+				<LeftAlignedRow>
 					{operatorDropdown}
-					<FilterContainer>
+					<IconButton
+						title={'Remove this criterion'}
+						iconProps={{ iconName: 'Delete' }}
```
darthtrevino, line 209: Spawn module variable: `deleteIconProps`

darthtrevino (approved):

javascript/react/src/controls/SortInstruction.tsx
natoverse: yeah, lodash merge considers the left-most object the "target", so it gets mutated. This reminds me I meant to move this out to a reusable style though since the same thing is used in a couple of places

PR #318, opened 2022-09-02
Schema updates

natoverse:


javascript/schema/src/datapackage/DataPackage.ts
```diff
@@ -12,5 +12,5 @@ import type { Resource } from './Resource.js'
  */
 export interface DataPackage extends Named {
 	$schema: string
-	resources: Resource[]
+	resources: (string | Resource)[]
```
darthtrevino, line 15: Can we add a typedoc comment to this field?

javascript/react/src/components/StepHistoryList.tsx
```diff
@@ -123,11 +123,14 @@ function onRenderHeader(step: Step): JSX.Element {
...
-		} else if (typeof columnList === 'object') {			
+		} else if (typeof columnList === 'object') {
 			columns = Object.values(columnList)?.join(', ')
 		}
 	} catch {
-		console.error('ColumnList type is not being currently supported', typeof columnList)
+		console.error(
+			'ColumnList type is not being currently supported',
+			typeof columnList,
+		)
```
darthtrevino, line 133: should we re-throw the error instead of swallowing it?

darthtrevino (approved):

javascript/react/src/components/StepHistoryList.tsx
natoverse: will note to Moni

PR #319, opened 2022-09-02
Feature/type discovery

gaudyb:


javascript/tables/src/guessDataTypeByColumn.ts
```diff
@@ -0,0 +1,83 @@
...
+		trueValues: trueDefaults,
+		falseValues: falseDefaults,
+		naValues: naDefaults,
+		dateFormat: dateFormatDefault,
+		thousands: thousandsDefault,
+		decimal: decimalDefault,
+		comment: commentDefault,
+	}
+
+	const guesser = guessDataType(defaultTypeHints)
```
natoverse, line 21: I think if you leave these defaults out they’ll be used anyway, right?

javascript/tables/src/guessDataTypeByColumn.ts
```diff
@@ -0,0 +1,83 @@
...
+	if (mapTypes.size === 1) {
+		for (const key of mapTypes.keys()) {
+			return key
+		}
+	}
+
+	return DataType.String
+}
+
+export const naDefaults = [
```
natoverse, line 50: Aren’t these redundant with the ones already defined in typeHints.defaults.ts?

natoverse (changes requested):

javascript/tables/src/generateCodebook.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Licensed under the MIT license. See LICENSE file in the project.
+ */
+import type { Codebook } from '@datashaper/schema'
+import type ColumnTable from 'arquero/dist/types/table/column-table'
+
+import { guessDataTypeByColumn } from './guessDataTypeByColumn.js'
+
+export function generateCodebook(table: ColumnTable): Codebook {
+	const codebookResult: Codebook = {
+		$schema: '???',
```
natoverse, line 12: https://microsoft.github.io/datashaper/schema/codebook/v1.json

javascript/tables/src/guessDataTypeByColumn.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function guessDataTypeByColumn(columnValues: string[]): DataType {
+	const guesser = guessDataType()
+
+	const mapTypes = new Map()
+
+	columnValues.forEach(value => {
+		const dataTypeResult = guesser(value)
+
+		if (dataTypeResult !== DataType.Null && !mapTypes.has(dataTypeResult)) {
```
natoverse, line 17: This seems overly complex. I would suggest iterating each value and simply putting the type in the map with `map.set`, including `DataType.Null` when the value is null/undefined/nan. At the end should be all the logic for determining the final type for the column: (1) delete DataType.Null, if there are no entries left the column type is null, (2) if there is more than one type remaining, the type is string, (3) otherwise return the remaining type

javascript/tables/src/guessDataTypeByColumn.ts
```diff
@@ -0,0 +1,37 @@
+/*!
+ * Copyright (c) Microsoft. All rights reserved.
+ * Licensed under the MIT license. See LICENSE file in the project.
+ */
+import { DataType } from '@datashaper/schema'
+
+import { guessDataType } from './guessDataType.js'
+
+export function guessDataTypeByColumn(columnValues: string[]): DataType {
```
natoverse, line 9: technically this doesn't care if it is a column and could be `guessDataTypeFromValues(values: string[])`

natoverse (approved):

PR #328, opened 2022-09-09
Add keepOriginalColumns flag to ONEHOT

natoverse:


javascript/react/src/verbs/OneHot.base.tsx
```diff
@@ -34,6 +34,13 @@ export const OneHotBase: React.FC<
 	const inputs = useMemo<FormInput<OnehotArgs>[]>(() => {
 		return [
 			inputColumnList(step, columns, 'Columns to onehot'),
+			{
+				label: 'Keep original columns',
+				type: FormInputType.Checkbox,
+				current: step.args.keepOriginalColumns,
```
darthtrevino, line 40: what argname are we using in other verbs? This seems somewhat verbose, maybe `preserveSource`?

darthtrevino (approved):

javascript/react/src/verbs/OneHot.base.tsx
natoverse: I agree. It's been that way for awhile but I was thinking of shrinking it. I like your suggestion

PR #331, opened 2022-09-13
papaparse and arquero readTable + tests

monik182:
Refactored the `readTable` method to use either papaparse or arquero depending on the parser options.
Created a bunch of utils to manage the parsers, map, and validate the options.


javascript/schema/src/datatable/ParserOptions.ts
```diff
@@ -56,5 +56,5 @@ export interface ParserOptions {
 	 * Number of rows to read from the data, starting at skip_rows (honors skipBlankLines).
 	 * Default: Infinity
 	 */
-	readRows?: number
+	nRows?: number
```
natoverse, line 59: I'm not really a big fan of `nRows`. I realize it is the arg name in pandas, but it feels to me like a formal JSONSchema should be a bit more verbose for clarity.

javascript/tables/src/readTable.utils.ts
```diff
@@ -39,21 +39,20 @@ export function getParser(options?: ParserOptions) {
...
-		table = from(subset)
-	}
-	return table
+	return options?.readRows ? from([...table].slice(0, options.readRows)) : table
 }
 
-function papaParser(text: string, options?: ParserOptions): ParseResult<any> {
+function papaParser(text: string, options?: ParserOptions): ColumnTable {
+	if (options?.skipRows && options?.readRows) {
+		options.readRows += options.skipRows
```
natoverse, line 47: I wouldn't mutate the actual input param. It's much safer to clone it first like `const opts = {...options}`. This may also result in an empty object if options is not present, which would remove the need for the `?` guards.

javascript/tables/src/readTable.utils.ts
```diff
@@ -39,21 +39,20 @@ export function getParser(options?: ParserOptions) {
 function arqueroParser(text: string, options?: ParserOptions): ColumnTable {
 	const mappedOptions = mapProps(ParserType.Arquero, options) as CSVParseOptions
 	let table = fromCSV(text, mappedOptions)
-	if (options?.nRows) {
-		const subset = nRows([...table], options.nRows)
-		table = from(subset)
-	}
-	return table
+	return options?.readRows ? from([...table].slice(0, options.readRows)) : table
```
natoverse, line 42: Is it necessary to copy this table array here with `[...table]`? Since it is a private variable you don't need to work about mutation.

natoverse: Oh, I think you are still adding more than needed here anyway - my previous comment should work where you create the table once and then return either a slice on it or the whole thing

natoverse (approved):

PR #338, opened 2022-09-16
Feature/remove min max bin

gaudyb:


python/datashaper/datashaper/engine/verbs/bin.py
```diff
@@ -38,19 +38,17 @@ def bin(
     to: str,
     column: str,
     strategy: str,
+    min: int,
```
natoverse, line 41: This may not be an int - it could be a floating point number

schema/fixtures/cases/verbs/bin/strategy_fixed_count/workflow.json
```diff
@@ -8,7 +8,9 @@
 				"to": "newColumn",
 				"column": "count",
 				"strategy": "fixed count",
-				"fixedcount": 5
+				"fixedcount": 5,
+				"min": 10,
```
natoverse, line 12: Can you add a test where the min/max are missing and see what happens?

natoverse (changes requested):

schema/fixtures/cases/verbs/bin/strategy_fixed_count/workflow.json
gaudyb: I have a couple of questions:
1. min/max should be optional? I saw this description in the BinArgs interface: **_If cell values are below this, they will default to -Infinity unless clamped_** from this I am assuming is not optional? Because right now what I did was args.min ?? Number.NEGATIVE_INFINITY and args.max ?? Number.POSITIVE_INFINITY. Or should I assume they are not optional but I should care they accomplish this rule **_If cell values are below this, they will default to -Infinity unless clamped_**
2. With the previous clarification I can also update the python version.

natoverse: I was thinking about it like the core Arquero bin function, which takes the parameters and computes a value's correct bin placement. They use +- inf as well https://github.com/uwdata/arquero/blob/main/src/op/functions/bin.js

The goal here was to remove any table stats compute from the verb, so the min/max are optional, but may not produce desirable results if they are not explicitly supplied. It would be good to test what the outputs are when they are not supplied to ensure it is a least sensible (i.e., does an infinite max break completely or result in millions of bins?).

Keep in mind that our user interfaces should suggest min/max values because it will have the metadata available (so it should start by suggesting the actual min/max from the table, which is what the verb used to do). So it should be an edge case to get a BIN instance that doesn't actually have a min/max (it would need to be hand-constructed JSON).

natoverse: New approach based on our conversation:
- revert the changes that remove the min/max calculation
- if min/max are supplied, do not perform `table.rollup` at all in the bin impl
- compute the min/max in the UI as defaults, so verbs created from the UI never have missing min/max (unless "auto" is specified)

natoverse (approved):

PR #340, opened 2022-09-21
Improve table api

natoverse:
- Cleans up a few of ArqueroDetailsList props that were excessive. Note that this is not yet addressing the need to remove the features in favor of a renderer-builder approach
- Removes the internal compute of table metadata, requiring it to be supplied if the table needs it (this was "auto magic" that should be explicit)
- Adds a bunch of stories for the table and header
- Migrates the examples from DebugPage to stories
- Removes DebugPage and router entirely, so the main webapp is just the prepare page


javascript/react/src/components/ArqueroDetailsList/__tests__/Columns/Columns.hooks.ts
```diff
@@ -0,0 +1,50 @@
...
+} {
+	const [checked, setChecked] = useState<string[]>(table?.columnNames() || [])
+	const checkboxes = useMemo(() => {
+		return (
+			table?.columnNames().map(column => ({
+				label: column,
+				checked: checked?.includes(column),
+				onChange: (
+					_ev?: React.FormEvent<HTMLElement | HTMLInputElement>,
+					chkd?: boolean,
```
darthtrevino, line 20: are we really saving that much by truncating this variable? Let's just call it `checked`

javascript/react/src/components/ArqueroDetailsList/__tests__/Columns/Columns.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useCheckboxConfigs, useColumns } from './Columns.hooks.js'
+
+export const Columns: React.FC<ArqueroDetailsListProps> = memo(
+	function Columns({ table, ...args }) {
+		const { checkboxes } = useCheckboxConfigs(table)
+
+		const { columns } = useColumns(checkboxes)
+
+		if (!table) {
+			return <div>Loading...</div>
```
darthtrevino, line 21: Hmm, this is fine for now but I wonder if we should think about including shimmer components for when things are in a loading-state.

javascript/react/src/components/ArqueroDetailsList/renderers/StatsColumnHeader.tsx
```diff
@@ -19,6 +19,7 @@ const pretty: Record<string, string> = {
 const CELL_HEIGHT = 14
 
 const DEFAULT_STATS: StatsColumnType[] = [
+	StatsColumnType.Type,
```
darthtrevino, line 22: What do Type and Invalid indicate here?

javascript/react/src/hooks/manageWorkflow.ts
```diff
@@ -126,11 +127,17 @@ export function useWorkflowOutputListener(
...
-				() =>
-					setOutput(workflow.toArray().filter(t => !!t) as TableContainer[]),
-				true,
-			),
+			workflow.onChange(() => {
+				const outputs = (
+					workflow.toArray().filter(t => !!t) as TableContainer[]
+				).map(table => {
+					if (!table.metadata) {
+						table.metadata = introspect(table.table!, true)
```
darthtrevino, line 135: Is this the right place to wire in introspection? This kind of implies that this hook _must_ be used to have metadata. I wonder if we should wire metadata into tables within the Worfklow class.

darthtrevino (approved):
I have some comments - we can chat about these today if you want

javascript/react/src/components/ArqueroDetailsList/__tests__/Columns/Columns.tsx
natoverse: It's just a story so I'm not super concerned

javascript/react/src/components/ArqueroDetailsList/renderers/StatsColumnHeader.tsx
natoverse: This is a enum of possible stats. Type shows the data type for the column, Invalid shows the count of invalid values (null, undefined, NaN)

javascript/react/src/components/ArqueroDetailsList/__tests__/Columns/Columns.hooks.ts
natoverse: Done

javascript/react/src/components/ArqueroDetailsList/ArqueroDetailsList.types.ts
```diff
@@ -143,16 +143,6 @@ export interface ArqueroDetailsListProps
...
-	 * Indicates whether we should include all of the columns in the table by default.
-	 * If false, a columns array must be provided.
-	 */
-	includeAllColumns?: boolean
-	/**
-	 * List of visible columns selected from the table.
-	 * If ommitted all configured columns are shown by default.
-	 */
-	visibleColumns?: string[]
 	isSortable?: boolean
```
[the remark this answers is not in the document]
natoverse: Did a bunch of these

javascript/react/src/hooks/manageWorkflow.ts
natoverse: It is not, but once we have our table-schema-generating flow we can opt in and inject it. In a case like this we may need some option to generate schemas automatically on derived tables, since it won't quite be in the user flow (like "Create codebook") is.

PR #350, opened 2022-09-23
removes input and output table from history list and fix lastStepOutput

dayesouza:
- history list without input and output table
- fix lastStepOutput when no steps
- enables saving in tableTransform when no mandatory output change
- removes unused property from ManageWorkflow component
- add history panel component to be reused
- removes add step button on history panel
- history panel layout fixes
- ordering for history panel with default as FirstOnTop


natoverse:
Some visual style comments:
- For the verb info button, it seems more standard for that icon to be at top right. You should also be able to shrink the width of the callout dialog even further.
- The history panel could be much narrower too (like 280px?)
- For the step order (#1, #2, etc.) it could be much lighter grey - this just needs to be a subtle indicator of ordering, not prominent visual information
- The step title does some sort of friendly camel-case of the column name, but that is confusing since it no longer matches the actual column name. I would (1) use the column name as is, and (2) when we have codebooks fully supported with the ability to specify a friendly name, support it there.
- Specify the default history order of "newest on top", which is reverse of current

javascript/webapp/src/pages/PrepareDataPage/PrepareDataPage.base.tsx
```diff
@@ -143,28 +133,12 @@ export const PrepareDataPage: React.FC = memo(function PrepareDataPage() {
...
-							<ManageWorkflow
-								workflow={workflow}
-								onSelect={setSelectedTableId}
-								onUpdateOutput={setOutputs}
-								onUpdateWorkflow={setWorkflow}
-								historyView={true}
-							/>
-						</WorkflowContainer>
-					</Aside>
+					<HistoryPanel
```
natoverse, line 136: Set title to "Steps"

javascript/react/src/components/HistoryPanel.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export const Aside = styled.div<{ isCollapsed: boolean }>`
+	display: flex;
+	flex-direction: column;
+	align-items: center;
+	height: 100%;
+	overflow: hidden;
+	box-shadow: ${({ isCollapsed }) =>
+		isCollapsed ? 'none' : '-1px 0px 10px 0px rgba(0,0,0,0.3)'};
```
natoverse, line 19: Don't use hard-coded colors: pull in the `theme` object (which should be injected in context and available as an arg automatically, see other styles below).

javascript/react/src/hooks/useWorkflowSteps.ts
```diff
@@ -3,19 +3,35 @@
...
+	workflow: Workflow,
+	order: WorkflowOrder,
+): Step[] {
 	const [steps, setSteps] = useState<Step[]>(workflow.steps)
 	// listen for workflow changes and update the steps
 	useEffect(() => {
-		setSteps(workflow.steps)
-		workflow.onChange(() => setSteps(workflow.steps))
-	}, [workflow, setSteps])
+		setSteps(orderSteps(workflow.steps, order))
```
natoverse, line 23: Why is `setSteps` called twice?

javascript/react/src/hooks/useWorkflowSteps.ts
```diff
@@ -3,19 +3,35 @@
...
+		workflow.onChange(() => {
+			setSteps(orderSteps(workflow.steps, order))
+		})
+	}, [workflow, order, setSteps])
+	return steps
+}
+
+function orderSteps(_steps: Step[], order: WorkflowOrder): Step[] {
+	let steps = cloneDeep(_steps)
```
natoverse, line 32: Do you need `cloneDeep` or can you just do something like `[..._steps]`?

natoverse (changes requested):

javascript/react/src/hooks/useWorkflowSteps.ts
natoverse: I guess you could go one step further too:
```
const steps = [..._steps]
return order === WorkflowOrder.FirstOnTop ? steps : steps.reverse()
```

javascript/react/src/components/ManageWorkflow.types.ts
```diff
@@ -2,11 +2,15 @@
  * Copyright (c) Microsoft. All rights reserved.
  * Licensed under the MIT license. See LICENSE file in the project.
  */
-import type { TableContainer } from '@datashaper/tables'
 import type { Workflow } from '@datashaper/workflow'
 
 import type { TransformModalProps } from '../index.js'
 
+export enum WorkflowOrder {
```
natoverse, line 9: This could actually be a much more generic enum for better reusability. Something like `OrdinalDisplay`.

javascript/react/src/components/ManageWorkflow.types.ts
```diff
@@ -2,11 +2,15 @@
...
+	FirstOnTop = 'first on top',
```
natoverse, line 10: Suggest not using spaces for enum values in case they're used as keys where spaces cause a problem. You can mash together `firstontop` or use dashes `first-on-top`

javascript/react/src/components/ManageWorkflow.types.ts
natoverse: Or the simpler `DisplayOrder` would be fine

javascript/react/src/components/HistoryPanel.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+	padding: 0;
+	margin: 0;
+	display: ${({ isCollapsed }) => (isCollapsed ? 'none' : 'block')};
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+`
+
+export const WorkflowContainer = styled.div<{ isCollapsed: boolean }>`
+	height: calc(100% - 75px);
```
natoverse, line 47: Can 100% work here?

natoverse: Point being that often a calc like this indicates a structural problem with the css

javascript/react/src/components/StepHistoryList.styles.ts
```diff
@@ -44,14 +47,16 @@ export const PanelHeader = styled.div`
 	gap: 0.5rem;
 	width: 95%;
```
natoverse, line 48: 95% is weird

javascript/react/src/components/StepHistoryList.styles.ts
```diff
@@ -3,6 +3,7 @@
  * Licensed under the MIT license. See LICENSE file in the project.
  */
 import styled from '@essex/styled-components'
+import type { Theme } from '@thematic/core'
 
 export const addButtonStyles = {
```
natoverse, line 8: Delete?

natoverse (changes requested):

natoverse:
Another style note: when the history panel is open and a verb is expanded, the delete icon seems to have more padding on the right than the eyeball on the left

javascript/react/src/hooks/useWorkflowSteps.ts
dayesouza: Yeahh that works, thanks!

javascript/react/src/components/HistoryPanel.styles.ts
dayesouza: Yep it works

natoverse (approved):
Looks awesome. One niggle - on prepare data the history still says "history" even though it is now "steps" in the flyout

PR #353, opened 2022-09-27
Steps count

dayesouza:
- removes manageWorkflow component
  - fixes a bug that wasn't updating the count when creating steps with the steps panel opened
- HistoryPanel expecting children
  - new showStepCount prop


javascript/react/src/components/HistoryPanel.types.ts
```diff
@@ -14,5 +14,6 @@ export interface HistoryPanelProps {
 	workflow: Workflow
 	title?: string
 	order?: DisplayOrder
+	stepsCount?: number
```
natoverse, line 17: This seems like a slightly weird expectation since the data is in the workflow. Maybe just a boolean like `showStepCount` and then still retrieve it from the workflow internally?

natoverse (changes requested):

natoverse (approved):

PR #356, opened 2022-09-28
Add Live Schema-Object Impls

darthtrevino:
* Constructors take optional schema objects
* All implement `toSchema`, `loadSchema`
* Add base classes for `Observed`, `Named`, and `Resource`


javascript/workflow/src/resources/DataPackage.ts
```diff
@@ -0,0 +1,117 @@
...
+		undefined,
+	)
+
+	private _codebook?: any
+	private _outputSubscription?: Subscription
+
+	public constructor(
+		id: string,
+		public readonly source: DataSource,
+		public readonly workflow = new Workflow(),
```
natoverse, line 35: Do you think this is sufficiently generic enough to handle any Resource type? It almost seems like we should have a list() and get(name) type of model?

natoverse: E.g., versus hard-coded workflow and codebook instances...

natoverse (approved):

javascript/workflow/src/resources/DataPackage.ts
darthtrevino: Well, the problem here is that in app-code we need to get the live instances, not just the generic schema variant. The data-package is kind of a weird outlier when it comes to serialize/deserialize from JSON because it's so abstract resource oriented.

natoverse: Maybe we can address this later with a plugin model. Ideally you'd have a factory function that you can register profiles with, which return live objects. So the data package won't really care about hard-coded instances - it will have a list of Resources, and retrieving one invokes the right factory, then passes it to the right component that knows how to render it...

PR #360, opened 2022-09-29
Python/Poetry Updates

darthtrevino:
* Use Poetry 1.2
* Use poethepoet for task definitions
* Include linter tasks

### DIFF
--- a/python/datashaper/datashaper/engine/verbs/unhot.py
+++ b/python/datashaper/datashaper/engine/verbs/unhot.py
@@ -25,15 +25,15 @@ def unhot(
         partial(__strategy_mapping[merge_strategy]), axis=1
     )
 
-    filteredList: list[str] = []
+    filtered_list: list[str] = []
 
     for col in output.columns:
         try:
             columns.index(col)
         except ValueError:
-            filteredList.append(col)
+            filtered_list.append(col)
 
     if not preserveSource:
-        output = output[filteredList]
+        output = output[filtered_list]
 
     return TableContainer(table=output)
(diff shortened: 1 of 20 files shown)


python/datashaper/datashaper/engine/verbs/utils/unhot_utils.py
```diff
@@ -6,15 +6,21 @@
 from datashaper.engine.verbs.verb_input import VerbInput
 
 
-def unhotOperation(input: VerbInput, columns: List[str], prefix: str):
+def unhot_operation(input: VerbInput, columns: List[str], prefix: str):
```
andresmor-ms, line 9: I know this PR is not about this, but was wondering if pandas melt operation would do this without the custom implementation? 

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.melt.html

python/datashaper/datashaper/engine/verbs/unhot.py
```diff
@@ -29,11 +29,11 @@ def unhot(
 
     for col in output.columns:
         try:
-            indexValue = columns.index(col)
+            columns.index(col)
         except ValueError:
             filteredList.append(col)
 
-    if preserveSource == False:
+    if not preserveSource:
```
andresmor-ms, line 36: now that you are renaming maybe rename this to `preserve_source`

darthtrevino: preserveSource is a little tricky because the JSONSchema uses camelcase. We'll need to convert from camel->snake as we read. It would be more idiomatic, but I don't think it's high priority.

python/datashaper/datashaper/engine/verbs/utils/unhot_utils.py
darthtrevino: That's a good thought, I'll add it as an issue

PR #366, opened 2022-10-04
add json shape to read table

dayesouza:


natoverse (commented):
Consider removing `export` here if you want to keep the public api simpler and just use this internally

darthtrevino (approved):

PR #369, opened 2022-10-05
readTable using props from arquero

dayesouza:


monik182 (approved):

javascript/tables/src/readTable.utils.ts
```diff
@@ -43,19 +43,19 @@ function arqueroParser(text: string, options: ParserOptions = {}): ColumnTable {
...
-	}
-	const mappedOptions = mapProps(ParserType.PapaParse, opts) as ParseConfig
-	const table = papa.parse(text, mappedOptions)
-	if (opts.skipRows) {
-		const subset = skipRows(table.data, opts.skipRows)
-		table.data = subset
-	}
-	return from(table.data)
-}
+// function papaParser(text: string, options: ParserOptions = {}): ColumnTable {
```
darthtrevino, line 46: what's going on here?

darthtrevino: there's a handful of commented out pieces - can we either delete these code blocks or add a comment about why they're commented out?

dayesouza: instead of using only fromCSV we need the whole mapping of params from arquero, so I had to re-enable readCsvTable,  but only use arqueroParser on the switch. By commenting the use of papaParser, lint complains that this function is unused if I leave it there alone

dayesouza: sure, I'll add a comment

PR #371, opened 2022-10-07
steps list change, fix creating same name step and update @essex/components

dayesouza:


javascript/react/src/components/StepHistoryList.types.ts
```diff
@@ -16,6 +16,7 @@ export interface StepHistoryListProps
 	order?: DisplayOrder
 	onDelete?: (index: number) => void
 	onDuplicateClicked?: (step: Step) => void
+	onSelectOriginalTable?: () => void
```
natoverse, line 19: Why does this need a different callback? I'm guessing because the workflow doesn't know if it's in the inputs list or outputs list, but it definitely feels weird to have to differentiate in this component's API.

javascript/react/src/components/TableCommands.tsx
```diff
@@ -63,11 +64,20 @@ export const TableCommands: React.FC<TableCommandsProps> = memo(
 					args: { to: selectedColumn, column: selectedColumn } as any,
 					input: inputTable?.id,
 				})
+				const id = createTableId(verb)
```
natoverse, line 67: Maybe do this before `readStep` and then include the id in the template? Seems weird to create and then mutate the id property

javascript/react/src/components/TableTransform.types.ts
```diff
@@ -83,7 +83,5 @@ export interface TableTransformProps {
 
 	onDelete?: (index: number) => void
 
-	onPreview?: (name: string) => void
-
 	onDuplicate?: (step: Step) => void
```
natoverse, line 86: Is this still necessary?

dayesouza: the onDuplicate? We're not using it now, should we remove it for good?

natoverse: Well, I see a couple of options to consider:
* Keep it as is, and ensure the functionality stays up-to-date to correctly duplicate the step. It would make sense to do the same with `onPreview` instead of removing it - but call it `onSelect` instead, which would give a consistent API to selecting tables.
* Remove it to simplify the API, and only have `onChange` and `onDelete` callbacks for the steps

I'm a little partial to option 2 because it means a simpler API for functionality that I don't think we'll need. One approach to getting the functionality back if it (or other functions) are desired would be to accept a command bar prop that replaces the default

natoverse: FWIW, I think the direct manipulation interface removes the need for step duplication - it's much more fluid, so I don't see why we'd need the duplicate button here. With an eventual advanced/graph editor maybe, but that'll be a different component...

javascript/react/src/components/StepHistoryList.types.ts
dayesouza: yep you're right. On the outside we're just getting the latest input anyway, so I've moved it inside

javascript/react/src/components/TableCommands.tsx
dayesouza: omg so obvious, yes thankss

javascript/react/src/components/TableTransform.types.ts
dayesouza: Yeah I think getting it back will be simple in case we need it, and clean to remove it for now

natoverse (approved):

PR #379, opened 2022-10-13
applyCodebook and unapplyCodebook

gaudyb:


javascript/tables/src/applyCodebook.ts
```diff
@@ -0,0 +1,38 @@
...
+	applyMapping: boolean,
+): ColumnTable {
+	const args = table.columnNames().reduce((acc, cur) => {
+		const field = codebook.fields.find(element => element.name === cur)
+		const parser = parseAs(field != null ? field.type : DataType.String)
+		acc[cur] = escape((d: any) => parser(d[cur]))
+		return acc
+	}, {} as Record<string, object>)
+
+	table = table.derive(args)
```
natoverse, line 24: Don't reassign input parameters. Instead, do something like `const applied = table.derive(args)`.

javascript/tables/src/applyCodebook.ts
```diff
@@ -0,0 +1,38 @@
...
+
+	if (applyMapping) {
+		const fieldList = codebook.fields.filter(element => element.mapping != null)
+		fieldList.map(field => {
```
natoverse, line 28: You should be able to use `fieldList.reduce` just like above and create a single set of `derive` args.

Then, if `applyMapping` is true, you can just `return applied.derive(args)` in the if block, and at the bottom of the function `return applied`.

javascript/tables/src/unapplyCodebook.ts
```diff
@@ -0,0 +1,43 @@
...
+		acc[cur] = escape((d: any) => parser(d[cur]))
+		return acc
+	}, {} as Record<string, object>)
+
+	table = table.derive(args)
+
+	if (applyMapping) {
+		const fieldList = codebook.fields.filter(element => element.mapping != null)
+
+		fieldList.map(field => {
```
natoverse, line 28: Same comment as in `applyCodebook`

javascript/tables/src/unapplyCodebook.ts
```diff
@@ -0,0 +1,43 @@
...
+	codebook: CodebookSchema,
+	applyMapping: boolean,
+): ColumnTable {
+	const args = table.columnNames().reduce((acc, cur) => {
+		const parser = parseAs(DataType.String)
+		acc[cur] = escape((d: any) => parser(d[cur]))
+		return acc
+	}, {} as Record<string, object>)
+
+	table = table.derive(args)
```
natoverse, line 23: Same comment as in `applyCodebook`

schema/workflow.json
```diff
@@ -1,4 +1,4 @@
-{
+ -m "{
```
natoverse, line 1: Is this what's causing the test to fail?

natoverse (changes requested):

PR #390, opened 2022-10-19
Remove Fetch, async graph nodes; refactor reactivity

darthtrevino:


natoverse (changes requested):
- remove fetch from top-level docs
- remove fetch impl from python

PR #404, opened 2022-10-26
column validation

gaudyb:


javascript/schema/src/codebook/ColumnError.ts
```diff
@@ -0,0 +1,8 @@
+/*!
+ * Copyright (c) Microsoft. All rights reserved.
+ * Licensed under the MIT license. See LICENSE file in the project.
+ */
+export interface ColumnError {
```
natoverse, line 5: It could be useful to add a list of cell indexes to this so we have the option of identifying the exact rows that have failures. We may not want to do that every time for performance reasons (see comment on validateTable), but when we do we could use it to highlight each cell that failed when viewing the table.

javascript/schema/src/codebook/ColumnError.ts
```diff
@@ -0,0 +1,8 @@
...
+	columnName: string
```
natoverse, line 6: You can just call this `name` for simplicity

javascript/schema/src/codebook/ValidationResult.ts
```diff
@@ -0,0 +1,9 @@
+/*!
+ * Copyright (c) Microsoft. All rights reserved.
+ * Licensed under the MIT license. See LICENSE file in the project.
+ */
+import type { ColumnError } from './ColumnError.js'
+
+export interface ValidationResult {
+	columnErrors?: ColumnError[]
```
natoverse, line 8: I would call this `errors` for simplicity (and you could reuse it elsewhere that we do validation)

javascript/schema/src/codebook/ColumnError.ts
natoverse: Hmm. An alternative idea here that may have better performance would be to just pass the validation results to the column when rendering (i.e., in ArqueroDetailsList). Then if the column has any rules violated we could test the cells as we render them. It might be more expensive in the render loop, but we'd also only need to test the visible cells with lazy loading.

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+export function validateTable(
+	table: ColumnTable,
+	codebook: CodebookSchema,
+): ValidationResult {
+	const validationResult: ValidationResult = {
+		columnErrors: [],
+	}
+
+	codebook.fields.forEach((field: Field) => {
+		const columnValues = table.array(field.name)
```
natoverse, line 23: You might want to confirm if a field has any constraints defined _before_ calling `table.array` to avoid lots of unnecessary table scans. I would just move the `if` statement below up outside of this.

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+	const columnError: ColumnError = {
+		columnName: name,
+		rule: message,
+	}
+
+	return columnError
+}
+
+function checkRequiredConstraint(values: string[]): boolean {
+	for (let i = 0; i < values.length; i++) {
```
natoverse, line 210: You can use `values.every` here for a more functional approach.

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+function checkUniqueConstraint(values: string[]): boolean {
+	const uniqueValues = [...new Set(values)]
+	return values.length === uniqueValues.length
+}
+
+function checkMinLengthConstraint(
+	minLength: number,
+	values: string[] | Array<any>[],
+): boolean {
+	for (let i = 0; i < values.length; i++) {
```
natoverse, line 231: You could use `values.every` here.

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	return true
+}
+
+function checkMaxLengthConstraint(
+	maxLength: number,
+	values: string[] | Array<any>[],
+): boolean {
+	for (let i = 0; i < values.length; i++) {
```
natoverse, line 242: `values.every`

natoverse: And the rest, I see there are several more loops so I'll stop adding comments :)

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+	table: ColumnTable,
+	codebook: CodebookSchema,
+): ValidationResult {
+	const validationResult: ValidationResult = {
+		columnErrors: [],
+	}
+
+	codebook.fields.forEach((field: Field) => {
+		const columnValues = table.array(field.name)
+		let flag = false
```
natoverse, line 24: I'm not sure what the best approach is with `flag` here.

- Right now you only get one error per column, because once `flag === true` you skip the rest of the checks.
- If you remove the flag, you can get a list of errors for the column, which could be helpful - at the expense of more perf required to check everything.

I think I am leaning toward the second option, because it would provide the user with a lot more information if there are problems with the table.

If you did that, you could also split this function apart quite a bit more like:

```
const requiredResult = checkRequiredConstant(columnValues as string, constraints) // returns a ColumnError or undefined
if (requiredResult) {
  validationResult.columnErrors.push(requiredResult)
}
```

That would move more of the logic into each checking function.

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+		if (field.constraints != null) {
+			const constraints = field.constraints
+
+			//required constraint
+			if (!flag && constraints.required != null && constraints.required) {
```
natoverse, line 29: I'm not sure why these are checking `!= null` _and_ falsy check. You should be able to do just the latter.

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+				const requiredResult = checkRequiredConstraint(columnValues as string[])
+				if (!requiredResult) {
+					flag = true
+					if (validationResult.columnErrors != null)
```
natoverse, line 33: You shouldn't need to check this each time - you initialized it to an empty array at the top of the function.

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+			if (
+				!flag &&
+				constraints.minLength != null &&
+				constraints.minLength &&
+				(field.type === DataType.Array || field.type === DataType.String)
+			) {
+				const minLengthResult = checkMinLengthConstraint(
+					constraints.minLength,
+					field.type === DataType.Array
+						? (columnValues as Array<any>[])
```
natoverse, line 63: I suspect you won't need to do this cast. Since you already checked if it is a string or array, and both of those types have a `length` property. So you can just simplify to `columnValues as any[] || string` I think.

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+						: field.type === DataType.String
+						? (columnValues as string[])
```
natoverse, line 65: This should be regular `string` - we want to ensure that the length of the string's characters meet the minimum, not test an array of strings (which is already covered by the array use case).

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+
+	return true
+}
+
+function checkMinimumConstraint(
+	minimum: number,
+	values: number[] | Date[],
+): boolean {
+	for (let i = 0; i < values.length; i++) {
+		if (values[i] != null && values[i]! < minimum) return false
```
natoverse, line 254: I'm not sure if this will work directly on Dates. You may need to call `valueOf` to get the numeric value, so you might want to split this into separate functions for the two different types.

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+
+	return true
+}
+
+function checkMaximumConstraint(
+	maximum: number,
+	values: number[] | Date[],
+): boolean {
+	for (let i = 0; i < values.length; i++) {
+		if (values[i] != null && values[i]! > maximum) return false
```
natoverse, line 265: Same as `checkMinimumConstraint`

natoverse (changes requested):

javascript/schema/src/codebook/ColumnError.ts
natoverse: You might even consider naming the interface `FieldError` or `ValidationError` to make it more reusable. One of the reasons I used `Field` in the codebook was because it may not always be a column in a table - it could be a property on a graph node for example.

javascript/tables/src/validateTable.ts
natoverse: This is because the `columnErrors` property is optional on `ValidationResult`. You can make it required if you will always be initializing it

javascript/tables/src/validateTable.ts
natoverse: Oh, right - I misread the cast - you are casting the entire list of values, so it is an array or arrays or array of strings.

javascript/tables/src/validateTable.ts
gaudyb: lint complains if you don't check first the nullability of the optional property and also the boolean

javascript/tables/src/validateTable.ts
gaudyb: Also it is an string[] and not string because the method receives the entire column list

javascript/schema/src/codebook/ErrorCode.ts
```diff
@@ -0,0 +1,14 @@
+/*!
+ * Copyright (c) Microsoft. All rights reserved.
+ * Licensed under the MIT license. See LICENSE file in the project.
+ */
+export enum ErrorCode {
+	Required = 'required constraint',
```
natoverse, line 6: I don't think you need to repeat the word "constraint" in all of these labels.

javascript/tables/src/validationUtils.ts
```diff
@@ -0,0 +1,324 @@
...
+				fail: !valuesCasted.every(validate),
+				indexes: resultIndexes,
+				rule: ErrorCode.Enum,
+			}
+		}
+	}
+}
+
+export function validateEnum(enumList: string[]): (value: string) => boolean {
+	return function validateEnum(value: string) {
```
natoverse, line 321: An optimization you can do here, which shows the cool power of closures: before returning the `validateEnum` function, you can add the `enumList` to a Set. This way the validation check can be `set.has` instead of `enumList.includes`, which will be much faster.

I.e.,

```
  const set = new Set(enumList)
  return function ... {
    return set.has(value)
  }
```

natoverse: The reason closure is cool here is because you only need to create the Set once, and then it is reused by the return function over and over instead of searching the list every time.

javascript/tables/src/validationUtils.ts
```diff
@@ -0,0 +1,324 @@
...
+				fail: resultIndexes.length != 0,
+				indexes: resultIndexes,
+				rule: ErrorCode.Required,
+			}
+		}
+	}
+}
+
+export function validateRequired(element: string): boolean {
+	if (element == null || element == undefined || element.length == 0)
```
natoverse, line 41: This function can be simplified to just `return (element === null || element === undefined || element.length === 0)`

Since you are returning a boolean either way, you can just return the result of the evaluation. Actually you need to invert the logic in this case, since you were returning false:

`return (element !== null && element !== undefined && element.length !== 0)`

Also note the extra `=` for strict checks.

One more question on this: it should apply for any data type, right, not just strings?

You could handle that by specifying the type as unknown and just doing the null/undefined check and ignoring the length. I think zero-length strings should be handled by the type application utility anyway (i.e., they will be turned to null already).

javascript/tables/src/validationUtils.ts
```diff
@@ -0,0 +1,324 @@
...
+		includeInstances: boolean,
+	): ValidationTestResult {
+		const uniqueValues = new Set<string>()
+		const resultIndexes: number[] = []
+		const valuesCasted = values as string[]
+
+		valuesCasted.map((value: string, index: number) => {
+			if (uniqueValues.has(value)) {
+				if (includeInstances) resultIndexes.push(index)
+			} else {
```
natoverse, line 62: This can be simplified a bit: since it is a Set, you can just add the value every time, so you don't need this `else` block.

javascript/tables/src/validationUtils.ts
```diff
@@ -0,0 +1,324 @@
...
+			})
+			return {
+				fail: resultIndexes.length !== 0,
+				indexes: resultIndexes,
+				rule: ErrorCode.Maximum,
+			}
+		} else {
+			return {
+				fail: !(dataType === DataType.Number
+					? (values as number[]).every(validate)
```
natoverse, line 233: Can you cast this to `(values as number[] || Date[])` to remove the logical switch?

javascript/tables/src/validationUtils.ts
```diff
@@ -0,0 +1,324 @@
...
+				indexes: resultIndexes,
+				rule: ErrorCode.Pattern,
+			}
+		}
+	}
+}
+
+export function validatePattern(pattern: string): (value: string) => boolean {
+	return function validatePattern(value: string) {
+		const re = new RegExp(pattern)
```
natoverse, line 285: Try moving the RegExp outside of the closure so it only needs to be created once. However, make sure this works ok because some RegExp methods are stateful. Whether you can do this depends on the usage (whether the RegExp is global).

javascript/tables/src/validateTable.ts
```diff
@@ -0,0 +1,288 @@
...
+						)
+				}
+			}
+		}
+	})
+
+	return validationResult
+}
+
+function createColumnErrorObject(name: string, message: string): ColumnError {
```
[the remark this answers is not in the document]
gaudyb: this code doesn't exist anymore, actually it says "Outdated" so I don't know why this is still part of the PR

javascript/tables/src/validationUtils.ts
gaudyb: I tried it but didn't work

natoverse (approved):

PR #407, opened 2022-10-28
Prune the React Surface Area

darthtrevino:
* Move internal-only components to an internal/ folder
* Remove project-management components not used in ShoWhy, move to webapp
* Remove vestigial components unused anywhere


natoverse (commented):
I think we can remove:
- RenameCallout. I don't see us supporting table rename from the header
- TableList/TableCard - these should all be dead?
- the export from `public.js` seems like a workaround for bad export * practices. what about if we only supply index.ts for the public surface area, and expect internal components to load an import directly?

darthtrevino:
> 

In this case the per-hooks imports can look pretty ugly. By having a single hook export we can unify all of our lib-hook imports into a single import statement per component.

 I'd be game for an `internal.js` file as an alternative that we can use for internal imports.

darthtrevino:
RenameCallout is used in the TableName component. TableCard and TableList have been removed

natoverse:
> RenameCallout is used in the TableName component. TableCard and TableList have been removed

Yeah, but we won't be allowing that so all of the associated logic can disappear

PR #427, opened 2022-11-15
Remove react-router-dom dep from app-framework

darthtrevino:
* Replace explicit react-router-dom with cascading props; this improves instrumentability and our ability to create stories
* Add basic storybook, only partial functionality atm
* Add `id`, `className`, and `style` props to DataShaperApp component


javascript/app-framework/src/components/DataShaperApp/DataShaperApp.types.ts
```diff
@@ -17,4 +34,14 @@ export interface DataShaperAppProps {
 	 * TODO: handle this internally
 	 */
 	children: React.ReactNode
+
+	/**
+	 * The currently selected route
+	 */
+	selectedRoute?: string
```
natoverse, line 41: Suggest naming as `selectedKey` to align with common control convention. Consumer doesn't necessarily know it is a "route" anyway

natoverse (approved):

PR #430, opened 2022-11-17
add codebook editor ui

dayesouza:


javascript/react/src/components/Codebook/CodebookTableEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+					newField,
+					...fields.slice(fieldIndex + 1),
+				])
+			},
+			[onChange, fields],
+		)
+
+		return (
+			<Container>
+				{fields.map((f: Field, index: number) => {
```
natoverse, line 33: You could create the onChange handler here and get the index property from the `map` operation, removing the need to find it above.

javascript/react/src/components/Codebook/MappingFields.hooks.ts
```diff
@@ -0,0 +1,68 @@
+/*!
+ * Copyright (c) Microsoft. All rights reserved.
+ * Licensed under the MIT license. See LICENSE file in the project.
+ */
+
+import { useCallback } from 'react'
+
+export function useChangeMappingValue(
```
natoverse, line 8: Can this be reused with Recode?

javascript/react/src/components/Codebook/MappingFields.tsx
```diff
@@ -0,0 +1,152 @@
...
+	FlexColumn,
+	MappingContainer,
+} from './MappingFields.styles.js'
+interface NewValueField {
+	value: string
+	display: string
+}
+export const MappingFields: React.FC<CodebookFieldEditorProps> = memo(
+	function MappingFields({ field, showInlineLabel, onChange }) {
+		const [values, setValues] = useState(field.mapping)
```
natoverse, line 27: Why do we need the internal state here?

javascript/react/src/components/Codebook/index.ts
```diff
@@ -0,0 +1,9 @@
+/*!
+ * Copyright (c) Microsoft. All rights reserved.
+ * Licensed under the MIT license. See LICENSE file in the project.
+ */
+
+export * from './CodebookFieldEditor.js'
```
natoverse, line 6: I suggest being explicit about each class and type we are exporting so we reduce the API area

javascript/react/src/components/Codebook/MappingFields.tsx
dayesouza: Different from recode, in here I always have at least one field and then I can fill it and click add, that adds the item to the table but we continue to show an empty field for the next one. But I will change it to be like recode where we only show the new field when clicked, then it will already be a part of the list anyway.

javascript/react/src/components/Codebook/CodebookFieldEditor.types.ts
```diff
@@ -0,0 +1,31 @@
...
+	Units = 'units',
+	Mapping = 'mapping',
+}
+
+export interface CodebookFieldEditorProps {
+	field: Field
+	onChange: (field: Field) => void
+	showInlineLabel?: boolean
+	showOutsideLabel?: boolean
+	showFields?: CodebookFields[]
```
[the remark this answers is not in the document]
dayesouza: This conflicts with the CodebookTableEditor where we receive a list of fields in the prop 'fields' :/ maybe we call this visibleFields then?

javascript/react/src/components/Codebook/CodebookFieldEditor.types.ts
```diff
@@ -0,0 +1,31 @@
...
+	DataType = 'data type',
+	DataNature = 'data nature',
+	Units = 'units',
+	Mapping = 'mapping',
+}
+
+export interface CodebookFieldEditorProps {
+	field: Field
+	onChange: (field: Field) => void
+	showInlineLabel?: boolean
```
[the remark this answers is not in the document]
dayesouza: I had to change it to hideLabel and tableView

javascript/react/src/components/ArqueroDetailsList/renderers/types.ts
```diff
@@ -62,9 +62,12 @@ export interface RichCellProps extends FormattedCellProps {
 	validationResult?: ValidationResult
 }
 
-export interface RichHeaderProps extends IDetailsColumnProps {
+export interface RichHeaderProps
+	extends Omit<IDetailsColumnProps, 'column' | 'columnIndex'> {
```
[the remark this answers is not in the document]
dayesouza: We weren't even using the extended properties, we just need an optional column, so I removed the extends and renamed it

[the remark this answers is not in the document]
natoverse: Hmm - I think we do want the extension from IDetailsColumnProps so our headers can pass them along to default renderers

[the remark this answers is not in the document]
dayesouza: ok, fixed using Partial

javascript/react/src/components/Codebook/CodebookFieldEditor.stories.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+const Template: ComponentStory<typeof CodebookFieldEditor> = ({
+	...args
+}): JSX.Element => {
+	const [field, setField] = useState(codebookResult.fields[0] as Field)
+
+	return <CodebookFieldEditor field={field} onChange={setField} {...args} />
+}
+
+export const Single = Template.bind({})
```
natoverse, line 139: If you only have one story and you name the export the same as the parent, Storybook will automatically roll it up to the parent node so you don't have to expand to find the child. In this example, naming const Single as CodebookFieldEditor would result in a single entry in Storybook tree, without the "Single" child you have to expand to.

javascript/react/src/components/Codebook/CodebookFieldEditor.types.ts
```diff
@@ -0,0 +1,31 @@
...
+export enum CodebookFields {
+	DisplayName = 'display name',
+	Description = 'description',
+	DataType = 'data type',
+	DataNature = 'data nature',
+	Units = 'units',
+	Mapping = 'mapping',
+}
+
+export interface CodebookFieldEditorProps {
```
natoverse, line 16: Add a styles block here so the user can customize whatever they need. There is a lot of hard-coded styling such as the widths, which may need overridden. essex-toolkit has a bunch of examples that I enabled deeper customization in so we could use them in DataShaper.

This would also provide a clean mechanism for editing the cell borders in single-versus-table mode. The default could be to have no cell border to reduce clutter, but have a cell style option, and the table editor could pass in the style to add the borders so it then looks like a table layout.

https://github.com/microsoft/essex-toolkit/blob/main/packages/components/src/ToolPanel/ToolPanel.types.ts

javascript/react/src/components/Codebook/CodebookFieldEditor.tsx
```diff
@@ -0,0 +1,149 @@
...
+						</FieldContainer>
+					</Then>
+				</If>
+				<If condition={visibleFieldsSet.has(CodebookFields.Description)}>
+					<Then>
+						{tableView && !hideLabel && (
+							<OutsideLabel>Description</OutsideLabel>
+						)}
+						<FieldContainer className="field">
+							<TextField
```
natoverse, line 72: Not sure about turning off the border on these controls. I understand the motivation, because the borders will cause it to look very cluttered. But it isn't obvious that the cell can be typed in. Maybe these need placeholder text? And maybe if it is in the entire table view we only put that placeholder text in the first column?

javascript/react/src/components/Codebook/CodebookFieldEditor.types.ts
```diff
@@ -0,0 +1,31 @@
...
+	field: Field
+	onChange: (field: Field) => void
+	tableView?: boolean
+	hideLabel?: boolean
```
natoverse, line 20: Here is an idea for getting rid of the `hideLabel` prop to clean up the interface:

1. Add the styles props as noted above, and include one for `label`.
2. Do not use the label prop on individual controls, instead add a <Label> or even just a div that is styled with the `styles.label` prop.
3. On the table editor view, set the label.styles display to 'none'.

This would eliminate all of the custom label logic you have through a single style prop.

javascript/react/src/components/Codebook/CodebookFieldEditor.tsx
natoverse: See my comments elsewhere about a styles block - this could be handled by a `styles.input` that can be used to turn the borders on or off. In table mode they're off (which is set by the table editor), in field mode they're on by default.

javascript/react/src/components/Codebook/CodebookFieldEditor.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+import { Label } from '@fluentui/react'
+
+export const FieldContainer = styled.div``
+export const OutsideLabel = styled(Label)`
+	position: absolute;
+	left: 3px;
+	margin-top: 10px;
+`
+
+export const Container = styled.div`
```
natoverse, line 16: General styling note: the fields definitely need some padding, which is apparent in the stats block and in single-field view where the labels bump right into the borders. I'd use the default padding we use on the verbs.

javascript/react/src/components/Codebook/CodebookFieldEditor.tsx
```diff
@@ -0,0 +1,149 @@
...
+								label={!tableView && !hideLabel ? 'Description' : undefined}
+								borderless
+								multiline
+								resizable={false}
+								disabled={field.exclude}
```
natoverse, line 77: This disabled doesn't seem to cover everything, so only part of the column goes grey. You may want to make even the stats header look disabled, and definitely make sure the mapping cell becomes disabled. It should look clear that the entire column is off, and the only way to get it back on is to re-check the box.

javascript/react/src/components/Codebook/CodebookFieldEditor.tsx
```diff
@@ -0,0 +1,149 @@
...
+								value={field.unit}
+								onChange={(_, val) => onChangeField({ unit: val })}
+							/>
+						</FieldContainer>
+					</Then>
+				</If>
+				<If condition={visibleFieldsSet.has(CodebookFields.Mapping)}>
+					<Then>
+						{tableView && !hideLabel && <OutsideLabel>Mapping</OutsideLabel>}
+						<MappingFields
```
natoverse, line 138: E.g., no `disabled` prop here...

natoverse (changes requested):

natoverse:
Another functional comment you might want to work into the story by setting the table container narrower than the total fields width: when horizontal scrolling, the left-side labels should stay fixed.

javascript/react/src/components/ArqueroDetailsList/renderers/StatsColumnHeader.tsx
```diff
@@ -32,7 +33,11 @@ export const StatsColumnHeader: React.FC<RichHeaderProps> = memo(
 				// data type is on the field, not the meta...
 				const value: any = stat === StatsColumnType.Type ? field.type : st[stat]
 				return (
-					<StatCell name={stat} value={value} key={`${column.key}-${stat}`} />
+					<StatCell
+						name={stat}
+						value={value}
+						key={`${column?.key || field.name}-${stat}`}
```
natoverse, line 39: IColumn.key should always be required, but since we're reusing this in a new context it isn't always there. This is a good indication to split this out into a better base component (and probably the histogram header too).

javascript/react/src/components/Codebook/Codebook.hooks.ts
```diff
@@ -0,0 +1,64 @@
...
+import {
+	DATA_NATURE,
+	DATA_TYPE,
+	DESCRIPTION,
+	DISPLAY_NAME,
+	MAPPING_FIELD,
+	MAPPING_WRAPPER,
+	STATS_WRAPPER,
+	UNITS,
+} from './Codebook.types.js'
```
natoverse, line 18: These would be more clear in a Codebook.constants.ts file

javascript/react/src/components/Codebook/Codebook.hooks.ts
```diff
@@ -0,0 +1,64 @@
...
+				setHeights(prev => ({
+					...prev,
+					[key]: value,
+				})),
+		}),
+		[heights, setHeights],
+	)
+}
+
+export function useUpdateMappingHeights(
```
natoverse, line 47: This whole hook seems like it should be a method on the FieldHeight object, like `updateAll`, and therefore contained in the `useFieldHeights` hook

javascript/react/src/components/Codebook/Codebook.styles.ts
```diff
@@ -0,0 +1,103 @@
...
+	styles?: CodebookDefaultStyles,
+	heights?: FieldHeights,
+): CodebookDefaultStyles {
+	const theme = useTheme()
+	return useMemo(
+		() => ({
+			mapping: {
+				root: {
+					height: heights?.get('mappingWrapper'),
+					...styles?.mapping?.root,
```
natoverse, line 36: once you start getting multiple nested merged, the lodash `merge` utility becomes a lot cleaner to use

javascript/react/src/components/Codebook/Codebook.styles.ts
```diff
@@ -0,0 +1,103 @@
...
+			root: {
+				width: 240,
+				flex: 'none',
+				border: `1px solid ${theme.palette.neutralTertiaryAlt}`,
+				borderBottom: 'unset',
+				...styles?.root,
+			},
+			displayName: {
+				root: {
+					padding: 10,
```
natoverse, line 49: Pull this padding into a const

javascript/react/src/components/Codebook/Codebook.styles.ts
```diff
@@ -0,0 +1,103 @@
...
+					height: heights?.get('displayName'),
+					borderBottom: `1px solid ${theme.palette.neutralTertiaryAlt}`,
+				},
+				...styles?.displayName,
+			} as ITextFieldStyles,
+			description: {
+				root: {
+					padding: 10,
+					height: heights?.get('description'),
+					borderBottom: `1px solid ${theme.palette.neutralTertiaryAlt}`,
```
natoverse, line 59: Create this border as a variable earlier so it can be reused

javascript/react/src/components/Codebook/Codebook.types.ts
```diff
@@ -0,0 +1,52 @@
...
+export const DISPLAY_NAME = 53
+export const DESCRIPTION = 83
+export const UNITS = 53
+export const DATA_TYPE = 53
+export const DATA_NATURE = 53
+export const STATS_WRAPPER = 127
+export const MAPPING_WRAPPER = 69
+export const MAPPING_FIELD = 32
+
+export interface CodebookDefaultStyles {
```
natoverse, line 30: you can remove `Default`

javascript/react/src/components/Codebook/Codebook.types.ts
```diff
@@ -0,0 +1,52 @@
...
+	mapping?: CodebookMappingStyles
+	statsWrapper?: CodebookStatsStyles
+	displayName?: ITextFieldStyles
+	description?: ITextFieldStyles
+	units?: ITextFieldStyles
+	dataType?: IDropdownStyles
+	dataNature?: IDropdownStyles
+}
+
+export interface CodebookDefaultProps {
```
natoverse, line 41: You can remove `Default`. If this is always a base, maybe call it `CodebookPropsBase`

javascript/react/src/components/Codebook/CodebookDescriptionField.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import type { CodebookTextFieldProps } from './Codebook.types.js'
+
+export const CodebookDescriptionField: React.FC<CodebookTextFieldProps> = memo(
+	function CodebookDescriptionField(props) {
+		const { styles, field, onChangeField, onChange } = props
+
+		const onChangeValue = useCallback(
+			(
+				_: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
```
natoverse, line 16: It's helpful to at least have some bit of name, even if just `_e` since it's an event

javascript/react/src/components/Codebook/CodebookTableEditor.tsx
```diff
@@ -0,0 +1,140 @@
...
+export const CodebookTableEditor: React.FC<CodebookTableEditorProps> = memo(
+	function CodebookTableEditor({ fields, onChangeFields, styles }) {
+		const heights = useFieldHeights()
+		const theme = useTheme()
+		const _styles = useTableDefaultStyles(styles, heights)
+
+		const onUpdateHeight = useUpdateMappingHeights(heights)
+
+		const onChangeField = useCallback(
+			(field: any, index: number) => {
```
natoverse, line 33: Remove `any`

natoverse (approved):
Looks like you got everything we talked about. My comments are stylistic.

javascript/react/src/components/ArqueroDetailsList/renderers/StatsColumnHeader.tsx
dayesouza: I'll create another PR for this change as we talked about

PR #433, opened 2022-11-21
Add Routing to DataShaperApp

darthtrevino:


javascript/app-framework/src/components/DataShaperApp/DataShaperApp.hooks.ts
```diff
@@ -51,20 +67,87 @@ export function useOnChangeWidth(
...
+			})
+
+			// Special case for raw resources embedded in datatable under `path` property
+			// and or datatable.json
+			if (isDataTableResource(resource)) {
+				if (resource.path != null && typeof resource.path === 'string') {
+					const pathItems = resource.path.split('/') ?? []
+					const lastPathItem = pathItems[pathItems.length - 1]
+					result.push({
+						path: `${root}/${resource.name}/${lastPathItem}`,
```
natoverse, line 123: I am a little torn on what the path should end up looking like. One consideration is that if the path contains things that are renamable by the user, bookmarks can eventually break. So we could use a guid that is generated for each resource. It also seems like we might want to consider adding some intermediary strings to indicate categories where relevant, like `${root}/resources/${resource.id}`?

javascript/app-framework/src/components/resources/FileTree/FileTree.types.ts
```diff
@@ -7,16 +7,48 @@ import type { DataShape } from '@datashaper/schema/dist/datatable/DataShape.js'
...
+
+	/**
+	 * Examples to present to the user in the command-bar menu
+	 */
+	examples: FileDefinition[]
+
+	/**
+	 * A map of profile-name to tree-node generator
+	 */
+	plugins: Map<string, DataShaperAppPlugin>
```
natoverse, line 41: Maybe not critical ATM, but it would be really nice to have a tree component where these types of plugin concerns are decomposed to custom render functions or contextual menu items, rather than domain-aware

javascript/app-framework/src/components/resources/FileTree/groupTables.ts
```diff
@@ -49,44 +38,35 @@ export function groupTables(tables: DataTable[]): ResourceTreeData[] {
...
+	const pathItems = (table.path as string).split('/')
+	const title =
+		pathItems[pathItems.length - 1] ?? `${table.name}.${table.format}`
 	return {
-		key: generateKey({
-			table: table.name,
-			resource: 'datasource',
-		}),
-		title: `${table.name}.${table.format}`,
+		href: `/resource/${table.name}/${title}`,
```
natoverse, line 45: This might obviate my prior comment about path structure partitioning, but it's a little hard to reason holistically about it in the PR viewer :)

natoverse (approved):
Looks pretty good so far

PR #434, opened 2022-11-22
Model Bundle Constraints Schema; Add 'tablebundle' Schema

darthtrevino:
* Add metaschema for defining data bundle constraint types
* Update base 'ResourceSchema' type so that the `sources` array may include relationship information via a `rel` attribute.
* Add 'tablebundle' schema type
* Add KnownProfiles enum to eliminate string literals for profiles
* Remove workflow, codebook chaining concerns from DataTable
* Move TableBundle persistence concern into a plugin
* Update DataShaperApp


javascript/schema/src/databundle/DataBundleSchema.ts
```diff
@@ -0,0 +1,31 @@
...
+ * This defines the data-bundle schema, which is a collection of table-centric resources.
+ * The data-table represents the data-source and parsing options.
+ * The codebook represents interpretation options and data-type mappings.
+ * The workflow represents data-wrangling steps.
+ */
+export interface DataBundleSchema extends ResourceSchema {
+	/**
+	 * Resource path to the datatable file, which contains the data source and parsing options.
+	 */
+	datatable?: string | DataTableSchema
```
natoverse, line 20: I don't think this is how we discussed the organization. Weren't we going to remove strongly-typed subschemas (the main one being codebook baked into DataTable I believe), and rely on resource handlers to understand how to apply their sources?

I thought the specific fixed types were going to end up as resource view config field wells, rather than on core schema

natoverse (approved):

PR #436, opened 2022-11-28
Feature/bin auto

gaudyb:


natoverse (commented):
Looks great so far

python/datashaper/datashaper/engine/verbs/bin.py
```diff
@@ -57,6 +57,30 @@ def bin(
         bin_edges = np.histogram_bin_edges(
             input_table[column], bins="auto", range=min_max
         )
+    elif bin_strategy == BinStrategy.Fd:
```
andresmor-ms, line 60: Now that there are so many strategies maybe you could change the pattern to avoid using all of these if/else.

I suggest creating a dictionary that use each of the `BinStrategy` as key and the value could be a lambda that performs the desired operation to get bin edges.

Something like:
```python
__bin_edges_mapping = {
    BinStrategy.Fd: lambda column, range: return np.histogram_bin_edges(column, bins="fd", range=range),
    # Other strategies
}
```

And then instead of all these branching, just use the map with the correct parameters

```python
bin_edges = __bin_edges_mapping[bin_strategy](input_table[column], min_max)
```

That way if in the future we need to add a new strategy we just add it to the map and that's it

natoverse (approved):

PR #437, opened 2022-11-29
Replace raw TreeView with Fluent Nav component

darthtrevino:


natoverse (approved):
Needs a bit of styling work, but this is really nice

PR #442, opened 2022-12-02
Fix Save-menu disabled bug

darthtrevino:


javascript/react/src/components/TableCommands/TableCommands.tsx
```diff
@@ -84,7 +84,7 @@ export const TableCommands: React.FC<TableCommandsProps> = memo(
 			],
 		)
 
-		const allTablesLengthObservable = useMemo(
+		const allTablesLengthObservable = useObservable(
```
natoverse, line 87: In the general react components it feels a little like Observable is an implementation detail that could be hidden underneath some collection of reusable hooks, e.g., `useTablesLength`. I realize this isn't a new change in this commit, it just became more obvious.

natoverse (approved):

javascript/react/src/components/TableCommands/TableCommands.tsx
darthtrevino: I can see the justification for that, yeah. I'd want to think through how to organize that.

PR #447, opened 2022-12-06
Streamline Menu Item Creation

darthtrevino:
Also, resolve issue where imported CSVs are created without a table-bundle.


javascript/app-framework/src/components/app/ResourcesPane/FileTreeCommands.hooks.ts
```diff
@@ -32,29 +40,122 @@ export function useFileManagementCommands(
...
-	const onClickUploadTable = useCallback(
-		() =>
-			void onOpenFileRequested(TABLE_TYPES)
-				.then(setFile)
-				.catch(err => console.error('error loading table file', err)),
-		[setFile, onOpenFileRequested],
+	return {
+		commands,
+		newCommands,
+		onOpenCommands: openCommands,
```
natoverse, line 93: The "on" prefix here makes this sound like an event hander - use `openCommands` and `saveCommands` to match `newCommands`, which would save the remap too?

natoverse (approved):

javascript/app-framework/src/components/app/ResourcesPane/FileTreeCommands.hooks.ts
darthtrevino: Agree

PR #452, opened 2022-12-14
Update tree

natoverse:
Brings in the latest essex Tree that support groups natively and per-item click handlers


javascript/app-framework/src/types.ts
```diff
@@ -140,7 +140,12 @@ export interface GeneratedExtraRoutes {
 	children?: ResourceRoute[]
 }
 
-export enum ResourceGroup {
+export enum ResourceGroupType {
```
darthtrevino, line 143: It's weird to me to add `Type` to the end of a type

darthtrevino: It's like starting an interface with `I` to me.

darthtrevino (approved):

javascript/app-framework/src/types.ts
natoverse: Yeah, but once I had a group with actual items in it (ResourceRouteGroup), it seemed weirder that one was an enum and the other was a structure. We can rename it if it bugs you.

PR #453, opened 2022-12-14
Generate codebooks when they are bound into table-bundles

darthtrevino:


javascript/workflow/src/resources/TableBundle.ts
```diff
@@ -264,4 +267,13 @@ export class TableBundle extends Resource implements TableEmitter {
 	): Maybe<TableContainer> => {
 		return table == null ? table : { ...table, id: this.name }
 	}
+
+	private populateCodebook() {
+		if (this.codebook?.fields.length === 0 && this.input?.output?.table) {
+			const name = this.codebook.name
+			this.codebook.loadSchema(generateCodebook(this.input.output?.table))
+			// generatecodebook will inject the name "generated"
```
natoverse, line 275: It probably shouldn't

natoverse (approved):

PR #463, opened 2022-12-29
Parser options

natoverse:
- split out the Parser component so it can be reused more easily
- reorg some of the editors


javascript/app-framework/src/components/editors/DataTableEditor/TableEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+					/>
+				</DetailsListContainer>
+				<ToolPanel
+					headerText={'Parser options'}
+					onDismiss={onToggleCollapsed}
+					headerIconProps={iconProps}
+					styles={toolPanelStyles}
+				>
+					<Parser
+						parser={resource?.parser?.toSchema()}
```
darthtrevino, line 65: toSchema() creates a new object every time - should we use `resource?.parser` instead so it can cache that object internally?

darthtrevino (approved):

PR #466, opened 2022-12-31
Load flows

natoverse:
- Fixes a bunch of the Import Table flows, enabling parameters that weren't working and supporting more JSON format options.
- Reorganizes the import dialog for better clarity and more options
- Uses a common component for updating parsing/format options in the TableEditor after load.


javascript/app-framework/src/components/DataTableConfig/DataTableConfig.hooks.ts
```diff
@@ -0,0 +1,42 @@
...
+export function useChangeHandlers(resource: DataTable): {
+	format: DataFormat
+	onChangeFormat: (
+		evt?: React.FormEvent<HTMLElement | HTMLInputElement>,
+		option?: IChoiceGroupOption,
+	) => void
+} {
+	const format = useUnobservedProperty<DataTable, DataFormat>(
+		resource,
+		'format',
```
darthtrevino: I see, `format` here is a sub-object which will generally be stable - why not use `resource.format`? If it's being mutated like this maybe we should make it observable.

javascript/app-framework/src/components/DataTableConfig/DataTableConfig.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+	...args
+}): JSX.Element => {
+	const resource = useMemo(
+		() =>
+			new DataTable({
+				format: DataFormat.CSV,
+			}),
+		[],
+	)
+	useEffect(() => resource.onChange(() => console.log(resource)), [resource])
```
darthtrevino, line 29: Maybe add `resource changed: `?

javascript/app-framework/src/components/DataTableConfig/Parser/Parser.tsx
```diff
@@ -0,0 +1,116 @@
...
+const lineTerminatorOptions = [
+	{ key: '\r', text: '\\r' },
+	{ key: '\n', text: '\\n' },
+	{ key: '\r\n', text: '\\r\\n' },
+]
+
+export const Parser: React.FC<ParserProps> = memo(function Parser({ parser }) {
+	const skipRows = useUnobservedProperty<ParserOptions, number>(
+		parser,
+		'skipRows',
```
darthtrevino: Is ParserOptions the main use case for this hook? Maybe we should just make these properties observables properly.

darthtrevino (approved):

PR #477, opened 2023-01-10
Metadata

natoverse:
Removes metadata from codebook


javascript/react/src/components/ArqueroDetailsList/ArqueroDetailsList.tsx
```diff
@@ -75,6 +77,8 @@ export const ArqueroDetailsList: React.FC<ArqueroDetailsListProps> = memo(
 			defaultSortDirection,
 		)
 
+		const _fields = useFields(table, fields)
```
darthtrevino, line 80: why the underscore here? That usually indicates a parameter variable that's being ignored

javascript/react/src/components/Codebook/CodebookFieldEditor/CodebookFieldEditor.stories.tsx
```diff
@@ -108,13 +58,77 @@ const codebookResult = {
 	],
 } as CodebookSchema
 
+const metadata: TableMetadata = {
```
darthtrevino, line 61: Do we have stories where the metadata is not explicitly defined like this, but is computed dynamically? It seems like we should support those cases. It would also be nice at some point if we could detect if the cached stats are invalid.

javascript/webapp/public/examples/smoking.json
```diff
@@ -16,2725 +16,333 @@
...
-									},
-									{
-										"min": 0.79955902103,
-										"count": 79
-									}
-								]
-							},
-							"nature": "continuous"
-						}
+					  {
```
darthtrevino, line 24: much nicer

darthtrevino (approved):

javascript/react/src/components/ArqueroDetailsList/ArqueroDetailsList.tsx
natoverse: We've used it in a variety of places inside of functions to indicate internal, but I agree it now conflates with ignored

javascript/react/src/components/Codebook/CodebookFieldEditor/CodebookFieldEditor.stories.tsx
natoverse: At one point we stopped computing dynamic metadata, because it would be recomputed every time a component got remounted (i.e., on nav), so we moved to expecting the app to provide it and manage that state independently

PR #484, opened 2023-01-17
Feature/bin nice param

gaudyb:


javascript/workflow/src/verbs/__tests__/binUtilities.spec.ts
```diff
@@ -124,4 +125,118 @@ describe('bin utilities tests', () => {
...
+
+		it('algorithm result', () => {
+			expect(min).toBe(10000000)
+			expect(max).toBe(80000000)
+			expect(width).toBe(10000000)
+		})
+	})
+
+	describe('regular numbers nice rounding test 1', () => {
+		const result = calculateNiceRounding(18, 67, 15.739759424941251)
```
natoverse, line 163: I think the minimum will always need to round down to make sure data values aren't excluded.

javascript/workflow/src/verbs/__tests__/binUtilities.spec.ts
```diff
@@ -124,4 +125,118 @@ describe('bin utilities tests', () => {
...
+
+		it('algorithm result', () => {
+			expect(min).toBe(0.1)
+			expect(max).toBe(1)
+			expect(width).toBe(0.196)
+		})
+	})
+
+	describe('small numbers nice rounding test 3', () => {
+		const result = calculateNiceRounding(0.1457, 0.9988, 0.15985224720734975)
```
darthtrevino, line 218: Be cautious of expecting high precision when using floats like this, precision may drift from machine to machine. You may want to do "close" checks for anything resulting from this instead of hard equality checks.

darthtrevino (approved):

PR #494, opened 2023-01-24
fix grouped list

dayesouza:
- groupedList is broken and there's a groupedListv2 that although is not in a stable release, has the grouped fixes

### DIFF
--- a/javascript/react/src/components/ArqueroDetailsList/GroupHeader.tsx
+++ b/javascript/react/src/components/ArqueroDetailsList/GroupHeader.tsx
@@ -3,19 +3,17 @@
  * Licensed under the MIT license. See LICENSE file in the project.
  */
 import type { IGroup } from '@fluentui/react'
-import { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { memo, useCallback, useMemo, useRef, useState } from 'react'
 import { Else, If, Then } from 'react-if'
 
 import {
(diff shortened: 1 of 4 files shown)


javascript/react/src/components/ArqueroDetailsList/GroupHeader.tsx
```diff
@@ -27,15 +25,8 @@ export const GroupHeader: React.FC<React.PropsWithChildren<GroupHeaderProps>> =
 		const [manualToggle, setManualToggle] = useState(false)
 
 		// trigger as soon as the element becomes visible
-		const inViewport = useIntersection(ref.current, '0px')
 		const countChildren = useCountChildren()
 
-		useEffect(() => {
```
dayesouza, line 33: @natoverse Do you think it's ok to remove this for now? It was always keeping the group opened even if we close it manually, so or we remove it or we put a conditional that it only auto-toggles on scroll if manualToggle=false

natoverse: Are you saying the that new grouped list behaves properly without this workaround?

dayesouza: Yes, the only downside is that when opening the 2 layered group it's not automatically opened

darthtrevino (approved):

PR #496, opened 2023-01-24
fix mapping in codebook so the newValue can accept any data type

dayesouza:


javascript/react/src/hooks/controls/useMappingPairs.tsx
```diff
@@ -107,15 +107,10 @@ const ColumnPair: React.FC<{
 				styles={{ root: { marginLeft: 4, marginRight: 4 } }}
 			/>
 
-			<DataTypeField
+			<TextField
```
natoverse, line 110: Are we losing the custom type entry stuff here?

dayesouza: Tor the value received yes, so the new mapped value can be anything. Shouldn't it?

natoverse: Another thought here: we have talked about adding a secondary codebook data type to indicate the new mapping “to” (maybe called “mappedType”). In that case we can just use a Freeform text box and just parse the string into that new type. It’ll be a less “dynamic” interface, but much simpler to deal with

natoverse: > Tor the value received yes, so the new mapped value can be anything. Shouldn't it?

In the DataTypeField, the idea was supposed to be that if the type is a Boolean, it’s just true/false, if it’s a number it’s a spinner, etc

natoverse: All we really need to do is make sure the data type for all mappings ends up the same since we can’t have mixed types in a column.  The lowest common denominator is a string

natoverse: Anything more dynamic is user convenience (and probably the only dynamic input that is really useful is the calendar picker for dates)

PR #506, opened 2023-02-08
Fix Companies Example; Workflow attach fix for Joins

darthtrevino:


javascript/webapp/public/examples/companies.json
```diff
@@ -30,7 +30,6 @@
 				{
 					"rel": "input",
 					"profile": "datatable",
-					"name": "companies",
 					"path": "data/products.csv"
```
natoverse, line 33: what name does this end up with? "datatable.json"?

darthtrevino: probably `datatable (3).json`

PR #507, opened 2023-02-09
Codebook edits

natoverse:
Fixes an issue where editing the codebook did not propagate to the table bundle viewer. This was essentially because table-impacting changes inside the codebook resource were not applying to recompute the output.

Also fixes a bunch of other bugs
- an issue where existing workflows could not be edited, because the steps weren't properly populating
- removes the outputs block from the workflow since it is essentially unused and all workflows currently only allow one "primary" output anyway
- fixes a couple of bugs in the binning (divide by zero and such)
- fixes the unroll verb to be a single column instead of a list
- fixes the join/lookup verb UX which were not populating the secondary table correctly
- removes Save button disabling for verbs, since many do not need changes and were therefore unsaveable
- adds codebooks for the companies example files so they look and work nicely now


javascript/workflow/src/resources/DataTable.ts
```diff
@@ -104,6 +88,20 @@ export class DataTable extends Resource implements TableEmitter {
 		return this._output$.value
 	}
 
+	private recomputeOutput = (): void => {
+		if (this._rawData != null) {
+			readTable(this._rawData, this.toSchema(), { autoType: false })
+				.then((t) => this._output$.next({ table: t, id: this.name }))
+				.catch((err) => {
+					log('error reading blob', err)
```
darthtrevino, line 96: we may want to push out undefined on the _output$ stream when an error happens

darthtrevino (approved):

PR #514, opened 2023-03-29
Patch Workflow to save output csv

natoverse:
This fixes an issue where the output of a workflow was never saved, and hence you could never retrieve the wrangled output. This fix works by adding a save function to the Workflow profile, and it also silently ignores any non-JSON resources when loading a saved zip (because the load mechanism was designed to only handle resource JSONs, and fails on any other file type). This is just a patch - eventually we need to support file types such as text or markdown by simply displaying them in the tree, but that will require more invasive work.


javascript/workflow/src/resources/DataPackage/LoadResourcesOperation.ts
```diff
@@ -206,22 +206,22 @@ export class LoadResourcesOperation {
...
+	private async readResource(
+		entry: string,
+	): Promise<ResourceSchema | undefined> {
 		// if the item is a string, look up the resource in the files map
 		let result: ResourceSchema | undefined
-
 		const blob = this.files.get(entry)
 		const resourceText = await blob?.text()
-		if (resourceText != null) {
+		if (entry.endsWith('json') && resourceText != null) {
```
darthtrevino, line 218: good catch here

javascript/workflow/src/resources/DataPackage/profiles/WorkflowProfile.ts
```diff
@@ -14,4 +14,18 @@ export class WorkflowProfile implements ProfileHandler<Workflow, WorkflowSchema>
...
 	}
+
+	public save(
+		data: Workflow,
+		dataPath: string,
+		files: Map<string, Blob>,
+	): Promise<string[]> {
+		const result: string[] = []
+		if (data.output?.table) {
+			const path = `${dataPath}output.csv`
```
darthtrevino, line 25: Maybe `${dataPath}${workflow.name ?? 'output'}.csv`?

darthtrevino (approved):

javascript/workflow/src/resources/DataPackage/profiles/WorkflowProfile.ts
natoverse: The downside here is that the workflow name is actually something like "workflow.json"

natoverse: Ok, latest commit fixes this by actually fixing a logic error I had previously made. The _workflow_ output is not what we want to save in the zip, it's the final _bundle_ output. So I moved the save function "up", and used the name, and the now matches the parent bundle name.

PR #524, opened 2023-05-04
Initial routes

natoverse:
- Updates initial route logic so avoid overriding existing navigation (e.g., initial route is only navigated if user is at root)
- Moves the app profile registration up the component hierarchy to avoid the duplicate registration warnings (AppResources and AppContent were both performing the same operation).


darthtrevino (approved):

javascript/app-framework/src/components/app/DataShaperApp/AppLayout.hooks.ts
```diff
@@ -274,13 +274,14 @@ export function useInitialDataPackageLoad(
...
-
+	const location = useLocation()
 	useEffect(
 		() => {
 			if (initialUrl != null) {
 				loadDataPackage(initialUrl)
 			}
-			if (initialRoute != null) {
+			// only navigate to the initial route if we're on the default homepage, otherwise respect prior navigation
+			if (initialRoute != null && location.key === 'default') {
```
darthtrevino, line 284: I'm looking at the docs for react-router-dom & history - it looks like "default" is used as a key on the initial render, before any history is pushed. 

I'm sure you tested this, but is there a risk that we're misusing the `key` property here?

darthtrevino (approved):

javascript/app-framework/src/components/app/DataShaperApp/AppLayout.hooks.ts
natoverse: Yeah, I set it up with the TestApp to confirm that it works correctly. Only needs to happen initially because that's when the initialRoute property overrides the root.

natoverse: The other option was a comparison with `pathname`, but that seemed more fraught (escape chars, etc.)

PR #528, opened 2023-06-09
Python API Extensibility; VAST Script

darthtrevino:
This PR introduces some changes to the Python API - namely:

* Set tables by id
* Initialize tables from dataframes in constructor
* Make JSON-schema validation optional (since the Python Json schema parser isn't working with our schema)
* Add verb extensibility
* Change `graph.get()` to `graph.output`
* Make string argument to `graph.output` optional to support default output.


python/examples/examples/vast.py
```diff
@@ -0,0 +1,89 @@
+from typing import List
+
+import pandas as pd
+
+from datashaper import ExecutionGraph, TableContainer, VerbInput
+
+dtype = {"date(yyyyMMddHHmmss)": "string"}
+vast1 = pd.read_csv("../data/vast/csv-1700-1830.csv", dtype=dtype)
```
natoverse, line 8: We should either bundle some sample data (with no redistribution issues) in the repo, or provide instructions for getting this data and locating it relative to the notebook

python/datashaper/datashaper/graph.py
```diff
@@ -42,42 +43,77 @@ class ExecutionGraph:
...
-            schema = json.load(schema_file)
-        validate(workflow, schema)
-
-        for input in workflow["input"]:
-            self.inputs[input] = TableContainer(
-                table=pd.read_csv(os.path.join(input_path, f"{input}.csv"))
-            )
-
+        # Perform JSON-schema validation
+        if validate:
```
natoverse, line 74: What happens here if validate is `True`, given the TODO comment above?

natoverse (approved):

python/datashaper/datashaper/graph.py
darthtrevino: It will blow up from the JsonSchema being invalid (there are some TS references in our schema - e.g. Record<string, something>, and Python doesn't appear to handle these well)

The validation is also problematic from the point of view that it's reaching outside of the library folder to find the default schema json file. In a publish version that would be a no-no.

python/examples/examples/vast.py
darthtrevino: I totally agree, I debated even putting it in this PR, but I think it demonstrates the API changes well and gives you, me and Alonso a strawman to work with

python/datashaper/datashaper/graph.py
darthtrevino: Validate seems to work in the tests - weird.

PR #529, opened 2023-06-22
Update UI to allow Arrow Import

darthtrevino:


javascript/tables/src/readTable.ts
```diff
@@ -33,8 +33,17 @@ export async function readTable(
 	}
 	const table = await textTable(input, schema)
 	const opts = defaultOptions(options)
+	const isTypedFormat = schema.format === DataFormat.ARROW
+
+	// TODO: should we eventually perform some casting/typing work on arrow tables? They're already strongly typed
```
natoverse, line 38: Maybe we should generate a codebook by reading the arrow typings, rather than inferring.

natoverse (approved):

javascript/tables/src/readTable.ts
darthtrevino: Hmm, maybe - I think that has to happen outside of readTable though. It takes a codebook in, but does not emit one as an artifact

natoverse: Yeah, it's in generateCodebook atm https://github.com/microsoft/datashaper/blob/main/javascript/tables/src/generateCodebook.ts

natoverse: `readTable` assumes you already have the schema, but on import if arrow is specified we can put alternate logic in `generateCodebook` to extract the types I suppose

PR #553, opened 2023-10-18
Add 'seed' arg to sample verb

darthtrevino:
Add 'seed' argument to the Sample verb.


AlonsoGuevara (approved):

natoverse (commented):
Since this is an existing verb with a UX counterpart, we should add a textbox to the form as well

natoverse:
Also note that, as your docs indicate, this may not apply in all environments - arquero does not appear to support a seed. Do we have a strategy for setting expectations/consistency?

AlonsoGuevara:
Does this change require any change in the JS library or in the integ tests?

darthtrevino:
@AlonsoGuevara the `sample` verb isn't tested, given it's random nature. It didn't make sense to set up a fixture test for it. It makes even less sense now that there's a little bit of a disconnect between JS and Python in this verb. We could unit test it though on the Python side.

darthtrevino:
@natoverse - Some verb documentation that highlights the JS/Python discrepancy is probably the best we can hope for until Arquero adds in a randomization seed argument.

natoverse:
> @AlonsoGuevara the `sample` verb isn't tested, given it's random nature. It didn't make sense to set up a fixture test for it. It makes even less sense now that there's a little bit of a disconnect between JS and Python in this verb. We could unit test it though on the Python side.

I don't think our fixtures support testing like this because they check for output file parity, but we could write a fixture that only confirms the expected output row count, based on the count/proportion args

darthtrevino:
FWIW @natoverse found a testing utility in Arquero we can use to set a seed value, eliminating the parity issue

PR #577, opened 2023-12-14
Verb decorator + Manager

andresmor-ms:


natoverse (approved):

python/datashaper/datashaper/engine/types.py
```diff
@@ -4,53 +4,10 @@
...
 
-from dataclasses import dataclass, field
 from enum import Enum
 from typing import Any, Union
 
-
-class Verb(Enum):
-    """The verbs that can be used in a pipeline."""
-
-    Aggregate = "aggregate"
```
darthtrevino, line 15: So nice

python/datashaper/datashaper/engine/verbs/unroll.py
```diff
@@ -3,10 +3,13 @@
 # Licensed under the MIT license. See LICENSE file in the project.
 #
 
+from datashaper.engine.verbs.verbs_mapping import verb
```
darthtrevino, line 6: I think all of these can be `from .verbs_mapping import verb`

darthtrevino (approved):

python/datashaper/datashaper/engine/verbs/unroll.py
andresmor-ms: I reallllly don't like relative imports xD maybe it's a personal thing, but after discussing with Rodrigo, in other projects, we usually do relative imports to resolve circular imports and in the `__init__.py` files and on the other places use absolute imports, what do you think? I'm working on doing this for datashaper project already.

AlonsoGuevara: +1 I personally prefer absolute imports. Relative imports are very obscure imo, in these things I prefer verbosity, to know exactly what I'm importing without needing to inspect.

python/datashaper/datashaper/engine/verbs/derive.py
```diff
@@ -36,6 +38,7 @@ def __concatenate(col1: pd.Series, col2: pd.Series):
 }
 
 
+@verb(name="derive")
```
AlonsoGuevara, line 41: Love this!!!

python/datashaper/datashaper/engine/verbs/__init__.py
```diff
@@ -1,2 +1,38 @@
+from . import strings  # noqa F401, F403
```
AlonsoGuevara, line 1: Wondering if there could be a way to avoid large inits with that many imports, this also happens on the pipeline

AlonsoGuevara (approved):

python/datashaper/datashaper/engine/verbs/__init__.py
andresmor-ms: `__init__.py` file is to expose the public API of a module, in recent versions of python (3.3+) you can just skip it if you don't want to define a public API for that module... In that case you just need to use the full import path (up to the file you need instead of the module). In this particular case we need to explicitly import the verbs so that the @verb decorator is executed, otherwise we would need to import all the verbs somewhere else.

natoverse: Could this be done lazily? Like, the verb registration is done on first lookup?

PR #580, opened 2023-12-19
Absolute imports

andresmor-ms:


natoverse (approved):

AlonsoGuevara (approved):

darthtrevino:
I prefer relative imports, but I'm willing to go along with this

AlonsoGuevara:
> I prefer relative imports, but I'm willing to go along with this

We should define what will be the standard moving fwd, as some of our projects (mainly the Ire-Indexing package) are using relative imports. In my case, I do prefer absolute, their verbosity makes it very clear what we are importing, in case like where we have things like "utils" or "prompts"

andresmor-ms:
From PEP-8:

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):
> 
> ```py
> import mypkg.sibling
> from mypkg import sibling
> from mypkg.sibling import example
> ```
> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:
> 
> ```py
> from . import sibling
> from .sibling import example
> ```
> 
> Standard library code should avoid complex package layouts and always use absolute imports.

https://peps.python.org/pep-0008/#imports


I think in python in general people prefer absolute imports and also like to keep the package structure simple enough to not require a giant import anyway. Also makes refactoring packages easier :). If you take a look at libraries like numpy or pandas they use absolute imports too.

PR #582, opened 2023-12-19
Function to load verbs instead of writing all the imports

andresmor-ms:
Proposal to have a function that loads all the verbs/**/*.py files to find `@verb` decorator and populate the VerbManager (this also removes the verbs from the public API and make them only accessible through the VerbManager)


AlonsoGuevara (approved):

natoverse:
Can you add a bit of documentation in the readme explaining where verbs are and how they need to mimic the package structure defined in the schema in order to be automatically found?

andresmor-ms:
> Can you add a bit of documentation in the readme explaining where verbs are and how they need to mimic the package structure defined in the schema in order to be automatically found?

Updated to include more docs about implementing new verbs in python

natoverse (approved):

PR #587, opened 2023-12-22
Improve verb loader function

andresmor-ms:
* Improve verb parameters to avoid using module properties as parameters


python/datashaper/datashaper/engine/verbs/__init__.py
```diff
@@ -11,11 +12,13 @@
 logger = logging.getLogger(__name__)
 
 
-def load_verbs(module_path: str, module_name: str):
+def load_verbs(module):
```
AlonsoGuevara, line 15: Nice!

AlonsoGuevara: So you pass a whole module? Or is it an object you created?

AlonsoGuevara (approved):

python/datashaper/datashaper/engine/verbs/__init__.py
andresmor-ms: You pass the whole module, there's an example in the vast notebook in the `examples` python project. In the case of the core verbs, I had to do `sys.modules[__name__]` to get the module loaded as a variable and pass it to the functions a few lines below.

PR #589, opened 2024-01-08
Minimize Full-Table Copies in Python Verbs to reduce Memory-Pressure

darthtrevino:
This PR introduces some topological analysis on top of Python-based workflows to minimize when we perform full-table copies. Due to the schema design of DataShaper, we've treated each verb as having an immutable output. This is great for flexibility, but in high-volume data scenarios this can result in memory pressure adding up quickly. 

In most of our use-cases, tables are simply handed off from one-verb to the next, and there's no need to maintain strict immutability rules except for a few rare cases. We did not want to present new "mutable" flags to verbs. Instead, this PR updates the Python engine so that it can determine when full-table copies are necessary. We use a couple of rules: 

* If the prior step has an explicitly defined "id", or if we're reading from a user-input table perform a full-copy
* If the current step declares that it is safe from input-table mutations, skip the full-copy.

We've added a new argument to the `@verb` annotation which allows verbs to declare themselves safe from mutation.

e.g. 

```py
@verb(name="aggregate", treats_input_tables_as_immutable=True)
def aggregate(...):
```

This begins the pattern of declaratively defining certain behavioral characteristics of verbs, which we anticipate will grow in the future with the need to consider compute-bound vs io-bound tasks and throttling across verbs on common semaphores.


python/datashaper/datashaper/engine/verbs/verbs_mapping.py
```diff
@@ -11,35 +11,62 @@
 from datashaper.table_store import TableContainer
 
 
-def verb(*args, **kwargs) -> Callable:
+def verb(name: str, does_not_mutate_input_tables: bool = False, **kwargs) -> Callable:
```
AlonsoGuevara, line 14: Generally one should avoid using negative variable names for booleans. I see you follow the guideline avoid using variables names that enable setting "false" by default.
What about a simpler name, less prone to generating confusion? :) preserve_input_table, keep_input_clean?

darthtrevino: That's a fair point, I renamed this to `treats_input_tables_as_immutable`

PR #597, opened 2024-01-16
Print JS and PY implementation

gaudyb:
- Print verb javascript implementation
- Print verb python implementation
- Workflow and jest tests


natoverse (changes requested):
Can you add one more optional arg to the verb, `limit`, with a default of 10? 10 would match arquero. I believe pandas `toString` may be the entire table, so by default for logging we will want a safe limit (calling `print(df)` defaults to 5 rows).

In arquero you can pass the `limit` arg as `limit` in the `table.print()` function, in pandas this is `max_rows` in the `df.toString()` function.

javascript/react/src/components/StepDescription/selectStepDescription.ts
```diff
@@ -61,6 +61,7 @@ const descriptions = {
 	join: JoinDescription,
 	lookup: LookupDescription,
 	merge: MergeDescription,
+	print: NoParametersDescription,
```
AlonsoGuevara, line 64: Should this be PrintDescription?

javascript/workflow/src/verbs/print.ts
```diff
@@ -0,0 +1,21 @@
+/*!
+ * Copyright (c) Microsoft. All rights reserved.
+ * Licensed under the MIT license. See LICENSE file in the project.
+ */
+import type { PrintArgs } from '@datashaper/schema'
+
+import type { ColumnTableStep } from './util/factories.js'
+import { stepVerbFactory } from './util/factories.js'
+
+export const DEFAULT_LIMIT_ROWS = 10
```
natoverse, line 10: You can put this in workflow/src/verbs/defaults so all default verb args are centralized

natoverse (changes requested):

natoverse (commented):
Did you commit the print.ts defaults file? I see the index export but not the file it references

natoverse:
> Did you commit the print.ts defaults file? I see the index export but not the file it references

Oh nevermind, maybe it was a previous commit

natoverse (approved):

PR #599, opened 2024-01-24
Adds ability to add callbacks to workflow

andresmor-ms:
* Adds callbacks to workflow (on workflow start/end and on step start/end)
* Adds an example to profile workflow memory usage and investigate allocations per verb and overall workflow.
* Bump datashaper version + update deps.
* Notebook showing the callbacks/profiling example: https://github.com/microsoft/datashaper/blob/mem_profiler/python/examples/notebooks/profiling.ipynb


AlonsoGuevara (approved):

python/examples/notebooks/vast.ipynb
```diff
@@ -63,6 +63,24 @@
    "execution_count": 3,
    "metadata": {},
    "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/home/andresmor/Projects/datashaper/python/examples/examples/verbs/custom_verbs.py:16: SettingWithCopyWarning: \n",
```
natoverse, line 70: Do you need to fix these paths?

natoverse (approved):

python/examples/notebooks/vast.ipynb
andresmor-ms: It's fine, those are warnings generated by pandas on some custom verbs

PR #605, opened 2024-01-31
Remove verb-timing callback, return verb-timings as Run Result

darthtrevino:


AlonsoGuevara:
General question, should we do all this changes on a new version of datashaper? 0.25?

darthtrevino:
> General question, should we do all this changes on a new version of datashaper? 0.25?

Yes, that will happen once this is at a point where we want to cut a release. When we change pyproject.toml's version number, once that hit main it will cut a release, and I want to avoid that for now

python/datashaper/datashaper/workflow.py
```diff
@@ -352,10 +354,14 @@ def run(
                 result = asyncio.run(result)
 
             # Record Verb Timing
-            if on_verb_timing is not None:
-                on_verb_timing(
-                    f"verb_{verb_name}_{verb_idx}", time.time() - start_verb_time
+            verb_timings.append(
```
andresmor-ms, line 357: are we always returning the timings? could be part of the @verb decorator and make the verb return timing, result

PR #607, opened 2024-01-31
Refactor Verb/Workflow Progress Callbacks

darthtrevino:
This PR accomplishes a few refactoring changes we discussed:

* Streamline the ad-hoc injection of callback functions into two main interfaces: `WorkflowCallbacks` and `VerbCallbacks`. Other status-reporting callbacks have been removed.
* Add an `options` argument to run, which can enable feature flags.
* Add a `memory_profile` feature flag to the workflow run, which will perform a memory profiling and inject those results into the workflow run results.
* Remove `StatusReporter` and its implementations: `FileStatusReporter`, `ConsoleStatusReporter`, `NoopStatusReporter`


python/datashaper/datashaper/workflow.py
```diff
@@ -304,55 +294,49 @@ def output(self, id: Optional[str] = None) -> Table:
     def run(
         self,
         context: Optional[Context] = None,
-        status_reporter: Optional[StatusReporter] = NoopStatusReporter(),
-        create_verb_progress_reporter: Optional[
-            Callable[[str], StatusReportHandler]
-        ] = _create_default_verb_status_reporter,
+        options: WorkflowOptions = WorkflowOptions(),
```
andresmor-ms, line 297: In python we usually avoid this, as this variable gets created when the class is loaded and not every time the method is called, this could cause confusion or errors, a way of avoiding this is to set it to None and then inside the method ask if it is None and set it to WorkflowOptions(),

python/datashaper/datashaper/types/verb_callbacks.py
```diff
@@ -0,0 +1,56 @@
+from typing import Any, Protocol
+
+from datashaper.progress.types import Progress
+from datashaper.types.workflow_callbacks import WorkflowCallbacks
+
+
+class VerbCallbacks(Protocol):
```
AlonsoGuevara, line 7: Nice! Protocols <3

python/datashaper/datashaper/workflow.py
AlonsoGuevara, line 297: rule of thumb is setting objects to None, and create them, if needed and set to None, in the method's body

python/datashaper/datashaper/types/workflow_results.py
```diff
@@ -0,0 +1,31 @@
...
+
+    id: str
+    index: int
+    verb: str
+    timing: float
+
+
+@dataclass
+class MemoryProfile:
+    snapshot_stats: pd.DataFrame
```
AlonsoGuevara, line 19: I think flake will complain about this missing docs?

python/datashaper/datashaper/types/workflow_options.py
```diff
@@ -0,0 +1,12 @@
+"""Data-types for workflow options and inputs."""
+
+from dataclasses import dataclass
+
+
+@dataclass
+class WorkflowOptions:
+    verbose: bool | None = None
```
AlonsoGuevara, line 8: Just curious, I was using verbose as the profiling flag. Now that it is separated, what do we envision verbose for?

python/datashaper/datashaper/types/verb_callbacks.py
```diff
@@ -0,0 +1,56 @@
...
+    def log(self, message: str, details: dict[str, Any] | None = None):
+        "Report an informational message from the verb execution."
+        ...
+
+    def measure(self, name: str, value: float):
+        "Report a telemetry measurement from the verb execution."
+        ...
+
+
+class DelegatingVerbCallbacks(VerbCallbacks):
```
AlonsoGuevara, line 31: With Protocols you don't need inheritance, the interpreter should be able to detect the proper protocol implementation.

darthtrevino: Right, so my thought here was that if a user was only interested in implementing a couple of workflow hooks, then they could subclass the base class and then define the specific functions they want. We could provide a protocol and a base class, but that seems like overkill to me

python/datashaper/datashaper/types/workflow_options.py
darthtrevino: Eh, good point, it's not used here yet, and could be feature creep. I'll remove it for now

python/datashaper/datashaper/types/workflow_results.py
darthtrevino: Good point, adding osme docs

python/datashaper/datashaper/types/verb_callbacks.py
darthtrevino: I went ahead and reverted to a protocol, but I added a new delegating implementation that guarantees all the methods are available.

PR #618, opened 2024-02-01
Initial commit draft caching tablestore

andresmor-ms:
Example notebook: https://github.com/microsoft/datashaper/blob/draft/caching_table_store/python/examples/notebooks/disk_cache_tablestore.ipynb


python/datashaper/datashaper/table_store.py
```diff
@@ -61,5 +67,118 @@ class TableContainer(Generic[T]):
...
+
+    def add(self, name: str, table: TableContainer) -> None:
+        """Add a table to the store."""
+        if self._verbose:
+            print(f"saving table {name} to disk")
+        table.table.to_parquet(f"{self._path}/{name}.parquet")
+
+    @cache
+    def _get_caching_function(self):
+        @lru_cache(maxsize=self._maxsize)
```
AlonsoGuevara, line 158: Nice

python/datashaper/datashaper/table_store.py
```diff
@@ -61,5 +67,118 @@ class TableContainer(Generic[T]):
...
+        def get_table(name: str) -> TableContainer:
+            if self._verbose:
+                print(f"loading table {name} from disk")
+            return TableContainer(pd.read_parquet(f"{self._path}/{name}.parquet"))
+
+        return get_table
+
+    def get(self, name: str) -> TableContainer:
+        """Get a table from the store."""
+        return self._get_caching_function()(name)
```
AlonsoGuevara, line 168: nit:
Could we hace the caching_function as a property, so you could so self._cache_get(name) ?

AlonsoGuevara:
So, to use this in the pipeline's workflows, we would need to prepend this to each workflow?

with DiskCacheTableStore(maxsize=5) as table_store:

Do you think there's a way you can configure this for a set of workflows? Like, making this shared?

andresmor-ms:
> So, to use this in the pipeline's workflows, we would need to prepend this to each workflow?
> 
> with DiskCacheTableStore(maxsize=5) as table_store:
> 
> Do you think there's a way you can configure this for a set of workflows? Like, making this shared?

You could do:
```py
with DiskCacheTableStore(maxsize=5) as tb:
    wf1 = Workflow(..., table_store=tb)
    wf1.run()
    wf2 = Workflow(..., table_store=tb)
    wf2.run()
    wf3 = Workflow(..., table_store=tb)
    wf3.run()
```

you can also do 

```py
table_store = DiskCacheTableStore(maxsize=5)
wf1 = Workflow(..., table_store=tb)
wf1.run()
wf2 = Workflow(..., table_store=tb)
wf2.run()
wf3 = Workflow(..., table_store=tb)
wf3.run()
```

The only caveat is that you would need to call to clear the files used manually

python/datashaper/datashaper/table_store.py
```diff
@@ -23,13 +29,13 @@ class ColumnStats:
...
     invalid: int
     mode: Any
-    min: Optional[float] = None
-    max: Optional[float] = None
-    mean: Optional[float] = None
-    median: Optional[float] = None
-    stdev: Optional[float] = None
-    bins: Optional[List[Bin]] = None
-    categories: Optional[List[Category]] = None
+    min: float | None = None
```
darthtrevino, line 32: I feel like we're going back and forth on these- I keep switching them to optional

python/datashaper/datashaper/table_store.py
```diff
@@ -47,19 +53,136 @@ class TableMetadata:
...
+        """List the tables in the store."""
+        pass
+
+    @abstractmethod
+    def clear(self) -> None:
+        """Clear the store."""
+        pass
+
+
+class DefaultTableStore(TableStore):
```
darthtrevino, line 105: Can we rename this to `MemoryTableStore`

darthtrevino (approved):

darthtrevino (changes requested):
1) We should add a `dispose` method that deletes the temp folder
2) We should discuss whether the 'max_size' is useful. If a table is generated at step 1 and used in step 99, we're actively introducing issues.

andresmor-ms:
> 1. We should add a `dispose` method that deletes the temp folder

The `clear` method already deletes the temp folder, however i agree dispose is a better name. 

> 2. We should discuss whether the 'max_size' is useful. If a table is generated at step 1 and used in step 99, we're actively introducing issues.

This is a LRU cache, that also saves the table into disk, when something is evicted because of the max_size of the cache and needed again it will be loaded from the parquet file on disk.


I'm also working on a TableStore that remove tables that are not needed anymore based on the dependency graph and one using diskcache python library.

darthtrevino (approved):

andresmor-ms:
Closing in favor of https://github.com/microsoft/datashaper/pull/639/files

PR #626, opened 2024-02-02
Parallel verbs decorator + example

andresmor-ms:
Verb decorator for parallel functions

Example: https://github.com/microsoft/datashaper/blob/feature/parallel_decorator/python/examples/notebooks/parallel.ipynb


python/datashaper/datashaper/engine/verbs/verbs_mapping.py
```diff
@@ -31,6 +43,98 @@ def inner(func: Callable[..., TableContainer]) -> Callable[..., TableContainer]:
...
+    return namedtuple(
+        old_named_tuple_type.__name__, old_named_tuple_type._fields + (new_column_name,)
+    )(*(old_tuple + (value,)))
+
+
+def parallel_verb(
+    name: str,
+    treats_input_tables_as_immutable: bool = False,
+    override_existing: bool = False,
+    asyncio_type: str = "asyncio",
```
darthtrevino, line 58: Can we make this a string enum? Let's start with "asyncio" and "asyncio_threads"

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
```diff
@@ -31,6 +43,98 @@ def inner(func: Callable[..., TableContainer]) -> Callable[..., TableContainer]:
...
+
+    def inner(func: Callable[..., pd.DataFrame]) -> Callable[..., TableContainer]:
+        @verb(
+            name=name,
+            treats_input_tables_as_immutable=treats_input_tables_as_immutable,
+            override_existing=override_existing,
+        )
+        async def wrapper(
+            input: VerbInput,
+            callbacks: Any,
```
darthtrevino, line 71: type is VerbCallbacks

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
```diff
@@ -31,6 +43,98 @@ def inner(func: Callable[..., TableContainer]) -> Callable[..., TableContainer]:
...
+
+            async def execute(
+                chunk: pd.DataFrame | tuple, tick: ProgressTicker, *args, **kwargs
+            ) -> pd.DataFrame | tuple:
+                try:
+                    output = await func(chunk=chunk, *args, **kwargs)
+                    tick(1)
+                    return output
+                except Exception as e:
+                    traceback.print_exc()
```
darthtrevino, line 107: let's not print the error for now - can we save the stack for the error reporting loop on line 122-123?

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
```diff
@@ -31,6 +43,98 @@ def inner(func: Callable[..., TableContainer]) -> Callable[..., TableContainer]:
...
+            *args,
+            **kwargs,
+        ):
+            input_table = input.source.table
+            chunk_size = kwargs.pop("chunk_size", 1)
+            if chunk_size == 1:
+                chunks = input_table.itertuples()
+            elif (
+                chunk_size > 0
+                and signature(func).parameters["chunk"].annotation == pd.DataFrame
```
darthtrevino, line 82: With datashaper we sometimes use our `Table` type alias - do you think this annotation mechanism is safe?

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
```diff
@@ -31,6 +43,98 @@ def inner(func: Callable[..., TableContainer]) -> Callable[..., TableContainer]:
...
+                try:
+                    output = await func(chunk=chunk, *args, **kwargs)
+                    tick(1)
+                    return output
+                except Exception as e:
+                    traceback.print_exc()
+                    errors.append(e)
+                    return None
+
+            futures = [
```
darthtrevino, line 111: Got to be careful here - this method looks like it calls execute() on every chunk by default, and fires off all of these async tasks concurrently. If these tasks hit throttled APIs out-of-the-gate we'll run into over-saturation quickly. We should semaphore the execute call or pass a semaphore into execute

darthtrevino (changes requested):
I think we're getting there, and I like this approach a lot, thanks for putting this together.

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
andresmor-ms: I know, but I was getting a circular dependency error, I'm going to take a look into this

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
darthtrevino: I take this back, leave the print, but add the stack trace to the callbacks below

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
andresmor-ms: I believe this will only create the futures but not execute them to actually execute them you need to call run, gather, or await it. From: https://docs.python.org/3/library/asyncio-task.html#coroutines

> Note that simply calling a coroutine will not schedule it to be executed

andresmor-ms: Oh I was using the semaphore in a wrong way :) fixing now

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
andresmor-ms: I changed it to Table... tbh I hate that type alias, this is just used because of the groupby verb in pandas returns a different type to be aggregated later, I'd like to see if we could either change the way this work or deprecate that verb since we already have the aggregate which does a groupby and aggregate back to a dataframe type.

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
andresmor-ms: I'll work this on another PR

AlonsoGuevara (approved):

python/datashaper/datashaper/engine/verbs/verbs_mapping.py
```diff
@@ -31,6 +44,120 @@ def inner(func: Callable[..., TableContainer]) -> Callable[..., TableContainer]:
...
+    """Enum for asyncio type."""
+
+    ASYNCIO = "asyncio"
+    THREADED = "threaded"
+
+
+def new_row(old_tuple: tuple, new_column_name: str, value: Any) -> tuple:
+    """Helper function to return rows in row-wise operations."""
+    old_named_tuple_type = type(old_tuple)
+    original_fields = old_named_tuple_type._fields  # type: ignore
```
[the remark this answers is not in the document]
andresmor-ms: _fields is a private field of namedtuple

darthtrevino (approved):

PR #630, opened 2024-02-07
Pyarrow/types

andresmor-ms:
* Updates code to be compatible with pyarrow types


darthtrevino (approved):

AlonsoGuevara (approved):
Curious about the experience. I see changes related to date times and to the unhot operation. Are these because of the migration? Or it was it seamless?

andresmor-ms:
> Curious about the experience. I see changes related to date times and to the unhot operation. Are these because of the migration? Or it was it seamless?

Because of the migration :) looks like some pandas functions don't take those types as datetimes so had to add another check. The unhot operation was failing because of type issues too, and from a long time ago I wanted to use `melt()` instead of a custom implementation for the unhot.

PR #632, opened 2024-02-08
Implement Workflow Verb

darthtrevino:
Implement the 'workflow' verb in Python to allow for sub-workflows. This verb is currently limited, in that sub-workflows can only accept a default input and emit a default output. Some more plumbing work in the datashaper internals will be required to arbitrary map-in input tables and draw-out output tables.


javascript/react/src/components/StepForm/selectStepForm.ts
```diff
@@ -94,7 +94,7 @@ const forms = {
 export function selectStepForm(
 	step: Step<unknown>,
 ): React.FC<StepFormProps<unknown>> {
-	const result: React.FC<StepFormProps<any>> = get(forms, step.verb)
+	const result: React.FC<StepFormProps<any>> | undefined = get(forms, step.verb)
```
natoverse, line 97: Is the expectation here that finding a workflow verb instance will throw on the client?

schema/fixtures/workflow/verbs/workflow/simple_inner_workflow/workflow.json
natoverse, line 1: Can you add a test that does more than one step? Also, potentially a recursive test where a workflow contains a workflow?

natoverse (commented):
We have a root docs folder for verbs that should get an entry for this.

For UX I think we can avoid building anything complicated right now, but it would be nice if it doesn't throw - maybe at least a VerbForm that just says "Worklow steps cannot be displayed" or something, so the app is stable and informative while incomplete.

javascript/react/src/components/StepForm/selectStepForm.ts
darthtrevino: The `get` interface can possibly return undefined, and the clause directly below this checks for null/undefined. 
This looks like a case of "I'm not sure how this passed CI to begin with"

schema/fixtures/workflow/verbs/workflow/simple_inner_workflow/workflow.json
darthtrevino: Sure thing - one limitation I'm seeing is that we can't get tables by `id` on an inner workflow step atm. I think we'd need to add multi-output nodes to get that capability.

python/datashaper/datashaper/workflow/workflow.py
```diff
@@ -234,16 +216,21 @@ def argument_names(fn: Callable) -> list[str]:
...
+        if (
+            "workflow_instance" in verb_args
+            and "workflow_instance" not in exec_node.args
+        ):
+            run_ctx["workflow_instance"] = self
+
         return run_ctx
 
-    def _check_inputs(self, node_key: str, visited: set[str]) -> bool:
+    def _missing_inputs(self, node_key: str, visited: set[str]) -> list[str]:
```
AlonsoGuevara, line 227: perhaps _get_missing_inputs (to start with an action verb)

schema/fixtures/workflow/verbs/workflow/simple_inner_workflow/workflow.json
natoverse: Given the implicit inputs when IDs are missing, is the chain something like:

`workflow input table -> workflow verb -> inner verb 1 -> inner verb 2 -> workflow verb output -> workflow output`

Is the expectation that the default workflow verb input is the last table before it, and the workflow verb output is the last inner verb's output? (i.e., it continues our normal non-workflow verb pattern).

And thus:

`workflow input table -> workflow verb -> inner workflow verb -> inner verb 1 -> inner workflow verb output -> workflow verb output -> workflow output` still results in a chained expected.csv that is easy to verify?

python/datashaper/datashaper/engine/verbs/workflow.py
```diff
@@ -0,0 +1,34 @@
...
+    input_arg: dict[str, str] | None = None,
+) -> TableContainer:
+    """Apply a sequence of operations to the input table."""
+    # Set up the input tables for the subworkflow
+    input_tables: dict[str, pd.DataFrame] = {
+        DEFAULT_INPUT_NAME: cast(pd.DataFrame, input.get_input())
+    }
+    if input_arg is not None:
+        for source, dest in input_arg.values():
+            input_tables[dest] = cast(pd.DataFrame, workflow_instance.output(source))
```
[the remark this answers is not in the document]
darthtrevino: We should definitely think through the `Table` typing at some point - it's either a pd.DataFrame or a pd.GroupBy currently

schema/fixtures/workflow/verbs/workflow/simple_inner_workflow/workflow.json
natoverse: Doing multiple inner workflows in a single steps[] and anything "parallel" gets much more difficult to verify because then you do need to grab tables by ID

javascript/react/src/components/StepForm/selectStepForm.ts
natoverse: Oh interesting, I thought you were explicitly supporting undefined as a return because there is no verb form for workflow. In the tagged verbs util we have a little test to ensure each verb has an entry; maybe we need something like this for the forms - theoretically `undefined` shouldn't be possible because that will invariably break the UI somewhere.

python/datashaper/datashaper/engine/verbs/workflow.py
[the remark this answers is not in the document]
natoverse: The difference is that the TableContainer has additional metadata fields that don't exist on the underlying tables

schema/fixtures/workflow/verbs/workflow/simple_inner_workflow/workflow.json
darthtrevino: The chaining you're describing is essentially correct, with a couple of notes:

* All steps in datashaper can reference other nodes as their main input- the prior is just the default
    * **roadmap item**: allow for nodes to have arbitrarily named inputs that they can use at their discretion
* All steps in datashaper can define an ID, which is usable in their immediate workflow.
* If you define steps IDs on an inner workflow, they're only usable in the inner workflow.
   * **roadmap item**: expose inner workflow IDs on the workflow node as named outputs
* If you have a workflow with two workflow nodes, and those nodes don't depend on each other - they can essentially run in parallel.

python/datashaper/datashaper/engine/verbs/workflow.py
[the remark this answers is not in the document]
darthtrevino: This is true, and we probably don't want to get rid of TableContainer, but the typing of `TableContainer.table` is causing downstream typing issues which necessitated the casting.

PR #633, opened 2024-02-09
Support different pandas types and backends

andresmor-ms:
* Make pyarrow optional due to some type issues (saving arrays to csv does not work correctly, also fails to save parquet files if arrays are used in columns, however it works if used as string)
* Extend tests to run over different dtype backend of pandas (numpy_nullable and pyarrow)


python/datashaper/datashaper/workflow/workflow.py
```diff
@@ -45,6 +46,13 @@
 DEFAULT_INPUT_NAME = "source"
 
 
+class PandasDtypeBackend(str, Enum):
```
AlonsoGuevara, line 49: What's the default?

andresmor-ms: `pandas_dtype_backend: PandasDtypeBackend = PandasDtypeBackend.NUMPY_NULLABLE,` in the constructor of the Workflow

AlonsoGuevara (approved):

python/datashaper/datashaper/workflow/workflow.py
AlonsoGuevara: Is this explicit in any docstring?

PR #639, opened 2024-02-12
Feature/table store

andresmor-ms:
* Refactor workflow to use TableStore as storage
* Provide in-memory and diskcache implementations


python/datashaper/datashaper/table_store/in_memory_table_store.py
```diff
@@ -0,0 +1,42 @@
...
+class InMemoryTableStore(TableStore):
+    """A default table store implementation."""
+
+    def __init__(self) -> None:
+        """Initialize the default table store."""
+        self._tables: dict[str, dict[str, TableContainer | str]] = {}
+
+    def add(self, name: str, table: TableContainer, tag: str = "default") -> None:
+        """Add a table to the store."""
+        self._tables[name] = {"table": table, "tag": tag}
```
darthtrevino, line 22: Can we use a small local dataclass instead of a dict?

darthtrevino (approved):

PR #640, opened 2024-02-12
Node Multi-Output

darthtrevino:
This PR updates the `step` concept with the ability to emit multiple named outputs, which will be useful in our graph applications where we want processing nodes that emit `nodes` and `edges`.


python/datashaper/datashaper/table_store/types.py
```diff
@@ -62,3 +62,11 @@ class TableContainer(Generic[T]):
     table: Table
     metadata: TableMetadata | None = None
     context: T | None = None
+
+
+@dataclass
+class ComplexVerbResult(Generic[T]):
```
natoverse, line 68: Does this actually need to be a separate concept from just a VerbResult? I.e., if we always allow multiple outputs, can we just skip the layer and consider that the standard option?

natoverse (commented):
Another note I didn't see in here: The NodeInput has always had "source" and "other", but not "others", not sure if we should revisit what should be in enum versus custom.

python/datashaper/datashaper/table_store/types.py
darthtrevino: I wanted to have minimal impact on existing verbs, which all emit 'TableContainer'. We could centralize this into 'VerbResult', but I'd rather do that as a follow-up PR

javascript/schema/src/workflow/verbs.ts
```diff
@@ -677,6 +677,11 @@ export interface SampleArgs {
 	 * The randomization seed to use for sampling to ensure stable sampling.
 	 */
 	seed?: number
+
+	/**
+	 * Whether to preserve and emit the non-sampled records via the 'unsampled' output port.
+	 */
+	emitUnsampled?: boolean
```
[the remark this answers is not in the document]
darthtrevino: This serves two purposes: 

* It gets us *just the rows* that were not sampled, which may be useful for training/test dataset splitting.
* It allows us to have a core verb with multiple output tables for integration tests.

javascript/react/src/components/StepList/StepHeader.utils.ts
```diff
@@ -48,7 +48,8 @@ export function deriveDetails(step: Step): string | undefined {
 
 	if (isInputColumnStep(step)) {
 		return (args as InputColumnArgs).column
-	} else if (isInputColumnListStep(step)) {
+	}
```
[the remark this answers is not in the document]
darthtrevino: It's just streamlining based on my linter. The `if` case returned, so there's not a need for an else

javascript/schema/src/workflow/verbs.ts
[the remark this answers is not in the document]
natoverse: Ok, on the first point I misunderstood the naming to mean the unsampled table, not the remainder rows. It might help to reword that slightly. Secondary suggestion though: is this actually needed as a param, or can the remainder just always be emitted as a secondary table?

javascript/react/src/components/StepList/StepHeader.utils.ts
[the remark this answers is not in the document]
natoverse: Hmm. I guess that's fine, but it does seem wonky that the linter enforces syntax that changes the logical expression (even if the practical result is the same).

javascript/schema/src/workflow/verbs.ts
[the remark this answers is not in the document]
darthtrevino: I like remainder, I think that's more descriptive. 

We _could_ always emit it, but it's more resource-intensive to do so

python/datashaper/datashaper/engine/verbs/sample.py
```diff
@@ -16,8 +18,24 @@ def sample(
...
     input_table = input.get_input()
-    output = input_table.sample(n=size, frac=proportion, random_state=seed)
-    return TableContainer(table=cast(Table, output))
+    if not emitRemainder:
+        output = input_table.sample(n=size, frac=proportion, random_state=seed)
+        return TableContainer(table=cast(Table, output))
+
+    result = input_table.sample(n=size, frac=proportion, random_state=seed)
+
+    sampled_ids: list = [row[0] for row in result.iterrows()]
```
andresmor-ms, line 31: if you want the sample and what was not selected you could do:

```py
result = input_table.sample(...)
remainder = input_table.drop(result.index)
```

python/datashaper/datashaper/workflow/workflow.py
```diff
@@ -284,9 +295,36 @@ def input_table(name: str) -> TableContainer:
...
-                input_mapping[key] = [input_table(t) for t in value]
-        return {"input": VerbInput(**cast(Any, input_mapping))}
+                raise WorkflowInvalidInputError(str(type(value)))
+
+        input_tbl = cast(TableContainer, input_mapping.pop("input", None))
+        source_tbl = cast(TableContainer, input_mapping.pop("source", None))
+        other_tbl = cast(TableContainer, input_mapping.pop("other", None))
+        others_tbl = cast(list[TableContainer], input_mapping.pop("others", None))
+        input_mapping = cast(dict, input_mapping)
+        return {
```
andresmor-ms, line 319: do we need to return a dict? can't we just return VerbInput?

python/datashaper/datashaper/engine/verbs/sample.py
andresmor-ms: Or: `unsampled = input_table[~input_table.index.isin(result.index)]`

python/datashaper/datashaper/table_store/types.py
darthtrevino: Renamed this to 'VerbResult' since the Complex part wasn't necessary. Verbs can still return a TableContainer, but all of our built-in verbs emit a VerbResult now.

PR #647, opened 2024-02-17
Destructure verb implementation

gaudyb:
This PR contains:

1. Javascript implementation for JSON object columns.
2. Python implementation with updates on test_schema.py file to support complex columns (JSON format).
3. UI changes to support this verb.
4. UI update that fixes a small render bug for objects on the table viewer.
5. Multiple workflow tests to test this functionality on both sides (PY and JS).
6. Markdown file with the verb explanation and examples.


javascript/react/src/components/StepDescription/selectStepDescription.ts
natoverse, line 1: Shouldn’t there be parameters to display in the description?

javascript/react/src/hooks/columns/useKeyNames.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function useKeyNames(
+	table: ColumnTable | undefined,
+	column: string,
+): string[] {
+	return useMemo(() => {
+		const distinctKeys = new Set<string>()
+
+		if (table !== undefined) {
+			const columnSelected: RowObject[] = table.select(column).objects()
```
natoverse, line 17: I would try to use native arquero functions as much as possible before switching to objects. For example, the op.keys can get the keys into an array https://uwdata.github.io/arquero/api/op#keys. I think you could then do an unroll verb, and finally and array_agg_distinct to get the unique list of keys for all rows. 

I’m not sure if this is faster, but usually arquero does a bunch of optimization internally that makes it much better than switching to the raw objects.

javascript/workflow/src/verbs/destructure.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	return preserveSource ? finalTable : finalTable.select(not(column))
+}
+
+function destructureSingleValue(
+	row: RowObject,
+	object: JSON,
+	keys?: string[],
+): RowObject {
+	for (let property in object) {
```
natoverse, line 39: Could something like this be useful? https://uwdata.github.io/arquero/api/op#entries

javascript/react/src/hooks/columns/useKeyNames.ts
gaudyb: I was able to replace the logic using op.keys but the op. array_agg_distinct did not work for me (you can see the updates on the PR)

natoverse (approved):

PR #654, opened 2024-02-20
node/output -> step/table

darthtrevino:


javascript/react/src/components/verbs/forms/LookupForm/LookupForm.base.tsx
```diff
@@ -44,7 +44,7 @@ export const LookupFormBase: React.FC<
 				getInputNode(step, NodeInput.Other),
 				(s, val) => {
 					const binding: InputBinding = (s.input[NodeInput.Other] as InputBinding) ?? { }
-					binding.node = val as WorkflowStepId
+					binding.step = val as WorkflowStepId
```
natoverse, line 47: Why is this different than JoinBase?

darthtrevino: Good catch, LookupForm.base is correct

PR #662, opened 2024-02-24
Housekeeping and bugs feb2024

natoverse:
This resolves a laundry list of issues, mostly related to UX consistency, parsing, and workflows.

- Fixes and issue with codebooks note respecting the "discover types" checkbox on import
- Adds a generic handler for unknown resources and simply displays the JSON. Previously the entire resource tree would be blank if it encountered unknowns
- Correctly navigates to new tables when opened
- Fixes default output column on step creation
- Updates the verb tagging to support more complex filtering, including strict data type checks
- The verb tagging lets us enable/disable individual verb buttons based on supported data types
- Cleans up the stories in several areas
- Enforces explicit step input ids on creation, which resolves a number of issues around alignment of tables with step form fields
- Fixes some issues with recent multi-output bindings refactor
- Filters table inputs to disallow things like joining a table to itself
- Cleans up naming of new/opened resources when there are potential collisions
- Respects Field title from codebook in ArqueroDetailsList, when provided
- Updates client-side boolean parsing in verbs and UX to use our core file parsing utilities, which aligns properly with pandas


javascript/app-framework/src/components/editors/TableBundleEditor/TableBundleEditor.hooks.ts
```diff
@@ -55,18 +62,32 @@ export function useColumnState(): [
...
 
+export function useCodebookFields(resource: TableBundle): Field[] | undefined {
+	const codebook = useMemo(() => {
+		const result = resource
+			.getSourcesWithProfile(KnownProfile.Codebook)
+			.find((t) => !!t) as Codebook | undefined
+		return result
+		/* eslint-disable-next-line react-hooks/exhaustive-deps */
+	}, [resource, resource.sources])
+	const fields$ = useMemo(() => {
```
darthtrevino, line 85: You don't really need the return - and since we're expecting a possible `undefined` then maybe `??` would be more appropriate? e.g.`useMemo(() => codebook?.fields$ ?? from([])`

javascript/react/src/components/ArqueroDetailsList/ArqueroDetailsList.hooks.ts
```diff
@@ -169,16 +169,19 @@ export function useListProps(version: number): IListProps<any> {
...
-		}
 		const types = columnTypes(table)
-		return Object.entries(types).map(([name, type]) => ({
-			name,
-			type,
-		}))
+		return Object.entries(types).map(([name, type]) => {
+			const existing = fields?.find((f) => f.name === name)
+			return (
+				existing || {
```
darthtrevino, line 179: existing ??

javascript/react/src/components/ArqueroDetailsList/renderers/functions.tsx
```diff
@@ -186,13 +186,16 @@ export const createRenderHistogramColumnHeader = (
...
-export const createRenderRowNumberColumn = (): IColumn => {
+export const createRenderRowNumberColumn = (rowCount = 1): IColumn => {
+	const minWidth = `${rowCount}`.length * 6
 	return {
 		key: ROW_NUMBER_COLUMN_NAME,
 		name: '',
 		fieldName: ROW_NUMBER_COLUMN_NAME,
-		minWidth: 20,
-		onRender: (_?: any, index?: number) => <RowNumberCell index={index} />,
+		minWidth: minWidth > 12 ? minWidth : 12,
```
darthtrevino, line 195: `Math.max(minWidth, 12)`?

darthtrevino (approved):

javascript/react/src/components/ArqueroDetailsList/ArqueroDetailsList.hooks.ts
natoverse: This is an old habit from before ?? existed. Are you aware of any lint rules that might catch these? I don't see anything on eslint other than support for the syntax.

darthtrevino: Not that I'm aware of - it's a pretty subtle difference between the two operators

natoverse: It is - though I suspect all of our assignments that use || could use ??

PR #669, opened 2024-03-13
Parallel filter

andresmor-ms:
Change the filter to be parallel verb in chunks


darthtrevino (approved):

darthtrevino:
This looks great - can you update the pyproject.toml file with a new version and cut a release with this?

PR #672, opened 2024-03-15
Add array parse type to convert verb in python

darthtrevino:


python/datashaper/datashaper/engine/verbs/convert.py
```diff
@@ -85,18 +86,32 @@ def _to_datetime(column: pd.Series) -> pd.Series:
...
 
+def _to_array(column: pd.Series, delimiter: str) -> pd.Series | pd.DataFrame:
+    def convert_value(value: Any) -> list:
+        if pd.isna(value):
+            return []
+        if isinstance(value, list):
+            return value
+        if isinstance(value, str):
+            return value.split(delimiter)
+        return []
```
AlonsoGuevara, line 97: Wonder if the default should instead be [value]

darthtrevino: Yeah good call

PR #673, opened 2024-03-15
Add array-input support to Python spread & aggregate verbs

darthtrevino:


python/datashaper/datashaper/engine/verbs/spread.py
```diff
@@ -3,75 +3,70 @@
...
 import pandas as pd
-from pandas._typing import Axes
 
 from datashaper.engine.verbs.verb_input import VerbInput
 from datashaper.engine.verbs.verbs_mapping import verb
 from datashaper.table_store.types import VerbResult, create_verb_result
 
+from .utils.type_utils import to_array_column
+
+_ONEHOT_ONLY_ACCEPTS_SINGLE_TO = "Onehot only accepts a single 'to' column."
```
natoverse, line 17: I believe this is incorrect - it should allow a list of columns, which limits the onehot columns

natoverse (approved):

python/datashaper/datashaper/engine/verbs/spread.py
darthtrevino: That seems flaky to me - it makes sense for the normal spread pattern, but with onehot the column-names are value-dependent. 

So if you were one-hotting values of X, would the column names indicate the column template or the values that you were interested in capturing? e.g. `['x_1', 'x_2', 'x_3']`? Would we need to split the input column names on `_` and check the second part for value quality after a string coercion? e.g. str(value) == column.split('_')[1]?

If not, how would you associate a value with a column?

natoverse: No, I think it is more like if you have array values of [“Red”, “Blue”, “Green”] but you only want output columns of Red and Blue. You’re just limiting the unique values that get onehotted.

natoverse: Note that this does not seem to be how the JS is implemented, so there’s still some general lack of clarity herr

natoverse: “to” seems to be entirely ignored in JS, which I don’t think was the original intent

natoverse: Let’s just put it this way: “to” is always being explicit about what the output columns are. If you have arrays with arbitrary lengths of 3 to 12 and specify 2 “to” columns, you only ever copy the first 2 values. With onehot you’re expressing the same explicit output

natoverse: Arquero’s “as”: https://uwdata.github.io/arquero/api/verbs#spread

PR #716, opened 2024-04-24
Verb defaults fixes

natoverse:
Fixes a few UX issues with verb input defaults or other weird glitches


javascript/react/src/components/verbs/forms/OrderbyForm/OrderbyForm.tsx
```diff
@@ -21,6 +21,25 @@ import { Container, icons, Sorts } from './OrderbyForm.styles.js'
...
+
+		useEffect(() => {
+			if (dataTable != null && onChange != null) {
+				if (!step.args.orders || step.args.orders.length === 0) {
+					onChange({
+						...step,
+						args: {
+							orders: [
+								{
+									column: dataTable.columnNames()[0] || '',
```
darthtrevino, line 33: Is catching a zero-column case here important?

darthtrevino (approved):

javascript/react/src/components/verbs/forms/OrderbyForm/OrderbyForm.tsx
natoverse: Sort of - the issue is that `column` is required for the verb per the schema, but a partial should be fine during UX building. We've always had a little bit of disconnect in various places like this where validation should happen, say, before allowing the user to save a step, but in the meantime _all_ properties on verb args could be Partial<T>

PR #724, opened 2024-05-01
New test structures

natoverse:
Adds support for JSON in cross-platform fixtures, which allows rewriting tests that were inadvertently obscuring issues in the verbs. This migrates a handful of tests to JSON to address some of those issues, but a deeper review is needed later. This also removes some date and object parsing "cheats" in the tests since JSON naturally handles those types better, and it removes a couple of tests that were invalid but masked by said cheats.


python/datashaper/tests/test_schema.py
```diff
@@ -25,42 +25,28 @@
...
         if file.endswith(".csv"):
-            path = Path(TABLE_STORE_PATH) / file
-            table = pd.read_csv(
-                path,
-                dtype_backend=pandas_dtype_backend,
-                engine=engine,
-            )
+            table = read_csv(path)
             dataframes[file.removesuffix(".csv")] = table
+        if file.endswith(".json") and file != "workflow.json":
```
darthtrevino, line 36: We're doing a lot of `endswith` in this PR - maybe it's worth a filename extension utility in datashaper/core or datashaper/tables?

schema/fixtures/workflow/verbs/destructure/plain_object_remove_column_no_keys/expected.json
```diff
@@ -0,0 +1,66 @@
+{
+    "ID": [
```
darthtrevino, line 2: the columnar json is a little ugly, but that's more of an aside than anything

darthtrevino (approved):

python/datashaper/tests/test_schema.py
```diff
@@ -25,42 +25,28 @@
 
 
 @cache
-def load_inputs(pandas_dtype_backend: PandasDtypeBackend, engine: str):
```
darthtrevino, line 28: So the issue here is that the tests aren't compatible with the Pyarrow backend anymore right?

natoverse: Correct - pyarrow requires jsonl, but regular engine (ujson) doesn't support it. I had to choose one, and default pandas made sense until pyarrow is 100% supported as an engine, and we may end up just using that for polars anyway.

schema/fixtures/workflow/verbs/destructure/plain_object_remove_column_no_keys/expected.json
natoverse: Agreed, but it is the standard for pandas and built-in to arquero so it kept the file loading minimal

python/datashaper/tests/test_schema.py
natoverse: Yeah, good point - we _do_ have that in the js impl, py is just way behind on file loading stuff

PR #744, opened 2024-05-30
Schema fixes

natoverse:
Couple of final schema cleanups after the larger reworking


darthtrevino (approved):
One side effect of this within the `filter` verb is that it precludes the capability to have rich boolean-expression filters. 

If you model each clause as a separate verb, you're essentially doing a giant "AND" clause, so we lose the expressiveness of being able to do OR operations. This may be perfectly reasonable, but it's something to consider.

natoverse:
> One side effect of this within the `filter` verb is that it precludes the capability to have rich boolean-expression filters.
> 
> If you model each clause as a separate verb, you're essentially doing a giant "AND" clause, so we lose the expressiveness of being able to do OR operations. This may be perfectly reasonable, but it's something to consider.

Absolutely correct. In the Loop component I put forward ways to achieve all of the removed functionality using individual verbs, and you can still model complex boolean logic - but it is much more painful now. You can do "OR" however by first performing individual column binarizes, then using the `boolean` verb to combine, and finally filter on that last output. In truth our boolean logic wasn't very flexible anyway since we had no notion of operator precedent or stacking - so it was "all or none" with the chosen operator, meaning you'd have to break out into individual verbs to model it anyway.

PR #747, opened 2024-06-03
Refactor: Streamline Verb Core Functions, Package Exports

darthtrevino:
The primary goal of this PR is to make verbs a first-class citizen and primary surface area of the Datashaper library. As such, the verbs have been moved from `datashaper.engine.verbs` to `datashaper.verbs`. 

Historically, our verb functions have been tightly coupled to the internal plumbing and I/O modeling of datashaper verbs (e.g. `VerbInput` input, `TableContainer` output). In order to support future use-cases in a more flexible manner, including being able to use plain verb functions as utilities in arbitrary Python code, we've introduced a series of monadic "Verb Adapter" decorators (namely `@inputs` and `@outputs`) designed to encapsule datashaper-specific boilerplate, allowing for verb-functions to be expressed in a more natural and generally useful manner.

The `@verb` annotation is considered to be _non-mutative_, any decoration or changes to the incoming function occur as side effects, and don't affect the signature or runtime behavior of the user-defined verb functions. Users may specify an `adapters` argument to declare how their user-defined functions should be utilized by datashaper.


python/datashaper/datashaper/constants.py
```diff
@@ -0,0 +1,8 @@
+#
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project.
+#
+"""Datashaper constants."""
+
+DEFAULT_INPUT_NAME = "source"
```
natoverse, line 7: Why not "input" and "output"?

python/datashaper/datashaper/utils/parallelization/types.py
```diff
@@ -0,0 +1,10 @@
+"""Execution utility typings."""
+
+from enum import Enum
+
+
+class AsyncType(str, Enum):
```
natoverse, line 6: I've been wary of "*Type" lately in python, and had to change some things in DataShaper to avoid conflicts (e.g., we had params like `type: ComparisonType`), which I've changed to somewhat equivalents like "*Strategy" (so now `strategy: ComparisonStrategy` for instance) to avoid weird potential built-in issues. Anyway, maybe worth considering on the name and subsequent usage here

python/datashaper/datashaper/verbs/copy.py
```diff
@@ -0,0 +1,32 @@
...
+    to: str,
+    column: str,
+    **_kwargs: Any,
+) -> pd.DataFrame:
+    """Copy verb implementation."""
+    table[to] = table[column]
+    return table
+
+
+apply_decorators(
```
natoverse, line 25: This feels like a layer of abstraction too far - just raw annotations on the verbs would be much more clear than sending to a wrapper function that probably doesn't actually reduce any lines of code

python/datashaper/datashaper/constants.py
darthtrevino: We could, it would require refactoring all of our fixtures though

python/datashaper/datashaper/verbs/copy.py
darthtrevino: Per our conversation, applying decorators as a separate expression like this allows us to export the base verb functions as standalone utilities. The @input and @output decorators fundamentally change these verbs for use with datashaper's verb semantics.

python/datashaper/datashaper/utils/parallelization/types.py
darthtrevino: That's fair, using another word here can probably let us be more accurate with some of these names. I've renamed a couple of these enums.

natoverse (approved):

